PR #2, opened 2025-04-28
Added new selector to reject galaxy-donuts

jfcrenshaw:


jmeyers314 (approved):
Looks good to me!

Can we take this opportunity though to add some logging to the selection code.  Something like:

self.log.info("%d of %d donuts passed power spectrum selection", sum(maxPowerGradSelect), len(maxPowerGradSelect))

jfcrenshaw:
> Can we take this opportunity though to add some logging to the selection code

Done!

PR #6, opened 2025-04-29
DM-50568: Parallelize EstimateZernikes

jmeyers314:
Adds an option to EstimateZernikesBase to parallelize over donuts.

On the summit, using 4 cores makes CalcZernikesTask run about 2x faster.


jbkalmbach (approved):
Looks good. Thanks!

PR #9, opened 2025-05-20
U/jbkalmbach/fix empty zernikes metadata

jbkalmbach:


suberlak (approved):
Looks good, it's a good idea to abstract adding metadata to its own function, both in calcZernikes and cutOutDonuts.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -327,45 +331,66 @@ def createZkTable(
...
-                "" if len(stamps) == 0 else stamps.metadata["BORESIGHT_ROT_ANGLE_RAD"]
+                stamps.metadata["BORESIGHT_ROT_ANGLE_RAD"]
             )
             dict_["boresight_par_angle_rad"] = (
-                "" if len(stamps) == 0 else stamps.metadata["BORESIGHT_PAR_ANGLE_RAD"]
+                stamps.metadata["BORESIGHT_PAR_ANGLE_RAD"]
             )
             dict_["boresight_alt_rad"] = (
-                "" if len(stamps) == 0 else stamps.metadata["BORESIGHT_ALT_RAD"]
+               stamps.metadata["BORESIGHT_ALT_RAD"]
```
suberlak, line 373: So if len(stamps)==0 , does this function still provide all metadata keys ?

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -327,45 +331,66 @@ def createZkTable(
...
-            (zkTable.meta["intra"], intraStamps),
-            (zkTable.meta["extra"], extraStamps),
+            (meta["intra"], self.stampsIntra),
+            (meta["extra"], self.stampsExtra),
         ]:
-            dict_["det_name"] = "" if len(stamps) == 0 else stamps.metadata["DET_NAME"]
-            dict_["visit"] = "" if len(stamps) == 0 else stamps.metadata["VISIT"]
-            dict_["dfc_dist"] = "" if len(stamps) == 0 else stamps.metadata["DFC_DIST"]
-            dict_["band"] = "" if len(stamps) == 0 else stamps.metadata["BANDPASS"]
+            if stamps is None:
```
suberlak, line 360: As this seems to imply that there won't be a `.metadata['DET_NAME']` for empty stamps...

suberlak (approved):

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: Added a test to verify this.

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: Should only come into play when we are using unpaired task.

PR #12, opened 2025-06-02
Add donut radius to donut quality table

gmegh:


python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -417,6 +412,9 @@ def empty(self, qualityTable=None, zernikeTable=None) -> pipeBase.Struct:
             "ENTROPY_SELECT",
             "SN_SELECT",
             "FINAL_SELECT",
+            "RADIUS",
+            "X_LEFT_EDGE",
```
jbkalmbach: Can you rename this and following "X_PIX_..._EDGE". Just to make clearer what X means.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -724,6 +736,20 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
             isEffective.append(eff)
             stampsEntropy.append(entro)
 
+            # Store donut radius in the metadata
```
jbkalmbach: This needs to be done differently. There should be no reason to set the configuration here when it gets it from the task. I think the way fitDountRadiusTask fit into the pipelines and the way it is currently written should be rethought.

jbkalmbach (changes requested):
Looking at how we are going to fit the donut radii and use them we should make some changes. 

1) `FitDonutRadiusTask` should be rewritten so that the `run` method runs on a list of donuts from the same side of focus and doesn't require both sides. That way you won't have to pass all the configurations back into it like you are currently doing.

2) Considering the way that we use `FitDonutRadiusTask`, I think that it should just be a normal task and not a pipeline task.

doc/versionHistory.rst
```diff
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-14.9.1:
```
jbkalmbach: Should also update to 14.10.0 if we make more extensive changes.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -724,6 +736,12 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
...
+            xLeft, xRight, radius, _ = self.fitDonutRadius.fit_radius(image=donutStamp.stamp_im.image.array)
```
jbkalmbach: I think you should move this down below and just use the `run` method on all the stamps at once.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -810,6 +828,11 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         stampsMetadata["EFFECTIVE"] = np.array(isEffective).astype(int)
         stampsMetadata["ENTROPY"] = np.array(stampsEntropy)
```
jbkalmbach, line 820: Just run the donut fit here once. Or alternatively, just change it so that `fit_radius` is the `run` method of the task and keep the structure here as it is.

jbkalmbach (approved):
Looks good! Just a final question on how best to use the task framework for fitting the radius.

PR #14, opened 2025-06-09
Add donut_radius fail flag to donutQualityTable

gmegh:


jbkalmbach (approved):
Looks good. Can you just add a check in the test for the task that it gives you the values you expect.

PR #15, opened 2025-06-09
Tickets/DM-51328: fix new schema in generateDonutFromRefitWcsTask

suberlak:


python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
```diff
@@ -216,14 +216,12 @@ def formatDonutCatalog(self, catalog):
         sourceSchema = afwTable.SourceTable.makeMinimalSchema()
         measBase.SingleFrameMeasurementTask(schema=sourceSchema)  # expand the schema
         # add coord_raErr,  coord_decErr to the schema
```
jbkalmbach: Extra whitespace after comma. Also, can you just add a little info on why we do this now.

doc/versionHistory.rst
```diff
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-14.11.0:
```
jbkalmbach: Should just be a minor update to 14.10.1 I think since this is just a fix of existing functionality.

jbkalmbach (approved):
Looks good, thanks for finding and fixing so quickly.

suberlak:
Also needed to update test values (<0.001 difference in SN calculated) in cutOutDonutsBase due to ISR updates in the weekly . Added rtol to be insensitive to such tiny changes next time.

PR #18, opened 2025-06-26
Tickets/dm 51473

jbkalmbach:


suberlak (approved):
Looks good, thanks for the  quick fix !

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -727,10 +727,6 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
             # Calculate the S/N per stamp
             snQuant.append(self.calculateSN(donutStamp))
 
-            # Add back in measured mean background to get mean background
```
suberlak: And that's not needed since we actually never removed background in `cutOutDonutsBase`, right? (I know we do +/-  background in `generateDonutsDirectDetect`  )

suberlak (approved):
Just one comment - do you keep track at which points we add / subtract background?

python/lsst/ts/wep/task/cutOutDonutsBase.py
jbkalmbach: We actually run the background task at the start of `cutOutStamps` which does subtract the background from the exposure that we use to cut out the donuts.

jbkalmbach:
We always run the background subtraction task. See here: https://github.com/lsst-ts/ts_wep/blob/c65a3f0636bc33e2e4d878e278802ddb97a74788/python/lsst/ts/wep/task/cutOutDonutsBase.py#L555

PR #19, opened 2025-06-27
Tickets/DM-51501: Mypy all ts_wep

gmegh:


.ts_pre_commit_config.yaml
```diff
@@ -1,9 +1,9 @@
-black: false
+black: true
 clang-format: false
-flake8: false
+flake8: true
 format-xmllint: true
-isort: false
-mypy: false
+isort: true
```
jbkalmbach: `black`, `isort` and `flake8` should still be false. `Ruff` is set up to handle all these now and that's why they were turned off.

jbkalmbach: See https://rubinobs.atlassian.net/browse/DM-50424

python/lsst/ts/wep/bin/ingestPairTable.py
```diff
@@ -71,7 +71,7 @@ def main():
     table = Table.read(args.pairs)
 
     logger.info(f"Writing pair table to butler collection {args.output_collection}")
-    butler = Butler(args.butler_config, writeable=True)
+    butler = Butler(args.butler_config, writeable=True)  # type: ignore
```
jbkalmbach: Why do you need an ignore on all the butler instances?

python/lsst/ts/wep/estimation/danish.py
```diff
@@ -475,28 +474,30 @@ def _estimatePairZk(
...
-                "variance": skyLevels[1],
-                "nollIndices": nollIndices.copy(),
-                "zkStart": zkStartI2.copy(),
-                "lstsqResult": result,
-                "zkFit": zkFit.copy(),
-                "zkSum": zkSum.copy(),
-                "model": modelImages[1],
-                "GalSimFFTSizeError": galSimFFTSizeError,
-            }
+            if I1 is not None and zkStartI1 is not None:
```
jbkalmbach: Isn't this bit unnecessary if `I1` and `zkStartI1` should `Image` and `np.ndarray` respectively? Unlike the `Z2` equivalents.

python/lsst/ts/wep/estimation/tie.py
```diff
@@ -22,7 +22,7 @@
 __all__ = ["TieAlgorithm"]
 
 import inspect
-from typing import Iterable, Optional, Union, Tuple
+from typing import Iterable, Optional, Tuple, Union
```
jbkalmbach: Why were you able to remove `Union` from `danish` but not from this?

python/lsst/ts/wep/estimation/tie.py
```diff
@@ -554,17 +554,21 @@ def _assignIntraExtra(
         if I1 is not None:
             if I1.defocalType == DefocalType.Intra:
                 intra = I1.copy()
-                zkStartIntra = zkStartI1.copy()
+                if zkStartI1 is not None:
```
jbkalmbach: `zkStartI1` should never be None so why is this necessary? Shouldn't you throw an error if you're worried about it being None?

python/lsst/ts/wep/estimation/tie.py
jbkalmbach: Same question for Optional

python/lsst/ts/wep/imageMapper.py
```diff
@@ -19,7 +19,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from typing import Optional, Tuple, Union
+from typing import Any, Optional, Tuple, Union
```
jbkalmbach: Same question about removing Optional and Union

python/lsst/ts/wep/imageMapper.py
```diff
@@ -1478,8 +1478,8 @@ def mapPupilToImage(
         image: Image,
         zkCoeff: Optional[np.ndarray] = None,
         nollIndices: Optional[np.ndarray] = None,
-        masks: Optional[Tuple[np.ndarray]] = None,
-        **maskKwargs,
+        masks: tuple[np.ndarray, np.ndarray, np.ndarray] | None = None,
```
jbkalmbach, line 1481: For instance here, why can masks remove `Optional` but not `zkCoeff` and `nollIndices`?

python/lsst/ts/wep/instrument.py
```diff
@@ -798,7 +802,7 @@ def getIntrinsicZernikes(
         xAngle: float,
         yAngle: float,
         band: Union[BandLabel, str] = BandLabel.REF,
```
jbkalmbach: More Unions in this file that could go away?

jbkalmbach (commented):
I'm working my way through but some comments thus far are included.

python/lsst/ts/wep/task/combineZernikesBase.py
```diff
@@ -43,11 +43,10 @@ class CombineZernikesBaseTask(pipeBase.Task, metaclass=abc.ABCMeta):
     ConfigClass = CombineZernikesBaseConfig
     _DefaultName = "combineZernikes"
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs: Any) -> None:
         pipeBase.Task.__init__(self, **kwargs)
-        self.log = logging.getLogger(type(self).__name__)
```
jbkalmbach, line 48: If we're removing this we should write a quick test that `combineZernikesSigmaClipTask` is still logging what we expect.

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
```diff
@@ -131,6 +133,8 @@ def runQuantum(
         so we put this in the dataId associated with the
         extra-focal detector.
         """
+        from lsst.ts.wep.task.generateDonutCatalogUtils import convertDictToVisitInfo
```
jbkalmbach: Why can't we keep this at the top?

python/lsst/ts/wep/task/reassignCwfsCutoutsTask.py
```diff
@@ -70,7 +72,7 @@ def adjust_all_quanta(self, adjuster):
                 seen.add(data_id)
             elif data_id["detector"] in intra_focal_ids:
                 extra_focal_data_id = DataCoordinate.standardize(
-                    data_id, detector=data_id["detector"] - 1
+                    data_id, detector=int(data_id["detector"]) - 1
```
jbkalmbach, line 75: Isn't `dataId['detector']` already an `int`?

python/lsst/ts/wep/utils/enumUtils.py
```diff
@@ -33,13 +33,13 @@
 
 from collections import UserDict
 from enum import Enum
-from typing import Any, Optional
+from typing import Any, Optional, Type
```
jbkalmbach: Same as earlier regarding `Optional`

python/lsst/ts/wep/utils/taskUtils.py
```diff
@@ -34,20 +34,26 @@
 import shlex
 import subprocess
 from contextlib import ExitStack
-from typing import Optional, Union
+from typing import Optional, TextIO, Union
```
jbkalmbach: `Optional`, `Union` once again

jbkalmbach (approved):
Looks good. A few questions that might just be me not too familiar with `mypy` yet. One request for a small test but shouldn't be complicated to implement.

python/lsst/ts/wep/task/reassignCwfsCutoutsTask.py
gmegh: I think that it's because even though you know data_id["detector"] is already an int, mypy can't infer this, especially when dealing with dictionary lookups, because from its perspective, dictionary values can have arbitrary types unless explicitly typed.

python/lsst/ts/wep/task/combineZernikesBase.py
gmegh: I have put that line back with ignore type

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
gmegh: Otherwise there is a circular import

gmegh: In particular, generateDonutCatalogUtils imports donutSourceSelectorTask. If we import that in cutoutDonutsScienceSensorTask, the donutSourceSelectorTask has not been imported yet given the init.py

python/lsst/ts/wep/bin/ingestPairTable.py
gmegh: Looks like a known issue: https://discovery-alliance.slack.com/archives/C2JPL8U78/p1729271936510669

I used from_config, to avoid ignoring the type

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
jbkalmbach: Let's move `convertDictToVisitInfo` to `lsst/ts/wep/utils/taskUtils.py` and move the test as well. Since it looks like we're not exclusively using it in `donutCatalog` tasks anyway that's probably the right move.

jbkalmbach: I think that will solve this issue.

PR #20, opened 2025-06-27
Tickets/DM-51516: make radius task more robust

suberlak:
Calculation of donut radius with median of cuts across different angles, on a postage stamp that may be binned. Adding appropriate config options to facilitate that: `nAngles` and `binning`

### DIFF
--- a/python/lsst/ts/wep/task/fitDonutRadiusTask.py
+++ b/python/lsst/ts/wep/task/fitDonutRadiusTask.py
@@ -3,9 +3,11 @@ import lsst.pipe.base as pipeBase
 import numpy as np
 from astropy.table import QTable
 from lsst.ts.wep.task.donutStamps import DonutStamps
+from lsst.ts.wep.utils import binArray
 from lsst.utils.timer import timeMethod
 from scipy.ndimage import gaussian_filter
 from scipy.signal import find_peaks
+from skimage.measure import profile_line
 
 __all__ = [
     "FitDonutRadiusTaskConfig",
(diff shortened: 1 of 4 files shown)


python/lsst/ts/wep/task/fitDonutRadiusTask.py
```diff
@@ -111,7 +125,9 @@ def run(
         failedFlagArray = []
 
         for stamp in stampSet:
-            xLeft, xRight, donutRadius, flag = self.fit_radius(stamp.stamp_im.image.array)
+            xLeft, xRight, donutRadius, flag = self.fit_radius(
```
gmegh, line 128: why are these lines changing? They look essentially the same as before

python/lsst/ts/wep/task/fitDonutRadiusTask.py
```diff
@@ -47,13 +49,23 @@ class FitDonutRadiusTaskConfig(pexConfig.Config):
...
     )
+    nAngles = pexConfig.Field[int](
+        doc="Number of angles for cross-sections.",
+        default=10,
+    )
+    binning = pexConfig.Field[int](
+        doc="Binning applied to each donut postage stamp.",
+        default=1,
+    )
+
```
gmegh, line 60: extra line?

gmegh (approved):
LGTM, please see my comments

python/lsst/ts/wep/task/fitDonutRadiusTask.py
suberlak: It was black (otherwise line was at >78, PEP8)

PR #22, opened 2025-07-23
Add background subtraction to generateDonutDirectDetectTask.

jbkalmbach:


gmegh:
Should we remove it then from the other step in CutoutDonutsBase?

jbkalmbach:
No because we don't actually save the background subtracted images back into the butler. It is a little bit of duplication but background subtraction is fast and if we move to a pure WCS catalog generation when the pointing is good we would eventually end up only doing it once.

suberlak (approved):
Thanks for the update, tested on 2025072100047  , helped with donut detection

PR #23, opened 2025-07-23
Tickets/DM-51659: add DonutCorrelator

suberlak:
This PR introduces DonutCorrelator to measure donut radii at the detection level. All metadata got updated to allow smooth integration with MTAOS.

### DIFF
--- a/python/lsst/ts/wep/donutSizeCorrelator.py
+++ b/python/lsst/ts/wep/donutSizeCorrelator.py
@@ -1,0 +1,342 @@
+import lsst.afw.image
+import numpy as np
+from lsst.ts.wep.utils import binArray
+from scipy.ndimage import binary_dilation
+from scipy.signal import correlate, find_peaks, peak_prominences
+
+
+class DonutSizeCorrelator:
+    """A class estimating donut diameters directly from post_isr_image.
+
+    This class estimates donut diameters by correlating the input image
+    with a series of annular templates at different diameters.
+
+    The main algorithm works by computing the maximum correlation between
+    each donut-sized template and the image, resulting in a curve of
+    maximum correlation as a function of donut diameter.
+
+    It then identifies the most prominent peak in this curve to estimate
+    the best-fitting diameter. If no strong peaks are found, the algorithm
+    falls back to analyzing the second derivative of the log-log curve
+    of diameter vs. maximum correlation to detect subtle inflection points.
+
+    To optimize performance, the algorithm uses a multi-scale strategy
+    with dynamically increasing resolution.
+
+    Example usage (assuming exposure is a post_isr_image object):
+        correlator = DonutSizeCorrelator()
+        image = correlator.prepButlerExposure(exposure)
+        diam, diamGrid, corrGrid = correlator.getDonutDiameter(image)
+        plt.plot(diamGrid, corrGrid)
+        plt.axvline(diam, c="r")
+    """
+
+    @staticmethod
+    def prepButlerExposure(
+        exposure: lsst.afw.image.Exposure,
+        badPixelTypes: tuple[str, str, str, str] = ("SAT", "BAD", "NO_DATA", "INTRP"),
+        nDilation: int = 100,
+        fillVal: float = -10.0,
+    ) -> np.ndarray:
+        """Prep image from the butler for correlation.
+
+        Parameters
+        ----------
+        exposure : lsst.afw.image.Exposure
+            Exposure from the butler.
+        badPixelTypes : tuple[str], optional
+            Names of flags that will be masked.
+            Default is ("SAT", "BAD", "NO_DATA", "INTRP")
+        nDilation : int, optional
+            Number of times to dilate bad-pixel mask.
+            Default is 100.
+        fillVal : float, optional
+            Value used to fill in mask. Warning, np.nan will
+            cause errors during correlation (FFT is used).
+            Default is -10.
+
+        Returns
+        -------
+        np.ndarray
+            The image array, prepped for correlation.
+        """
+        # Extract image and mask arrays
+        image = exposure.image.array
+
+        # Subtract off approximate background from image
+        image -= np.median(image)
+
+        # Get bad-pixel mask
+        bitMask = exposure.mask.getPlaneBitMask(badPixelTypes)
+        badPixels = np.bitwise_and(exposure.mask.array, bitMask) > 0
+
+        # Dilate the bad-pixel mask
+        mask = binary_dilation(badPixels, iterations=nDilation)
+
+        # Fill masked values
+        image[mask] = fillVal
+
+        # Normalize the image
+        image /= image[~mask].max()
+        return image
+
+    @staticmethod
+    def cropAndBinImage(
+        image: np.ndarray,
+        length: int | None = None,
+        pad: int = 500,
+        binning: int | None = None,
+    ) -> np.ndarray:
+        """Crop and bin the array.
+
+        Parameters
+        ----------
+        image : np.ndarray
+            Image array.
+        length : int or None, optional
+            Size length for crop. Default is None.
+        pad : int, optional
+            The image is cropped around the brightest pixel, subject
+            to the condition that the pixel is not too close to the
+            edge of the image. This pad sets that distance.
+            Default is 500 pixels.
+        binning : int or None, optional
+            Binning factor. Default is None.
+
+        Returns
+        -------
+        np.ndarray
+            Cropped and binned image
+        """
+        # Crop array
+        if length is not None:
+            # We will crop around the brightest pixel
+            # Cutout pads on side so we don't select for
+            # bright donuts falling off the sensor
+            padded = image[pad:-pad, pad:-pad]
+
+            # Bin the array so we don't select individual hot pixels
+            binningFactor = 8
+            binned = binArray(padded, binningFactor)
+
+            # Find location of brightest pixel
+            y, x = np.where(binned == np.nanmax(binned))
+
+            # Undo binning and pad
+            x = binningFactor * x[0] + pad
+            y = binningFactor * y[0] + pad
+
+            # Crop around brightest pixel
+            height, width = image.shape
+            x0 = np.clip(x - length // 2, 0, width - length)
+            y0 = np.clip(y - length // 2, 0, height - length)
+            image = image[y0 : y0 + length, x0 : x0 + length]
+
+        if binning is not None:
+            image = binArray(image, binning, "median")
+        return image
+
+    @staticmethod
+    def createDonutTemplate(diameter: float) -> np.ndarray:
+        """Create simple annulus template for donuts.
+
+        Parameters
+        ----------
+        diameter : float
+            Diameter of donut.
+
+        Returns
+        -------
+        np.ndarray
+            Fractional donut mask
+        """
+        # Create grid of pixel centers
+        x = np.arange(diameter + 5, dtype=float)
+        x -= x.mean()
+        x, y = np.meshgrid(x, x)
+
+        # Build sub-pixel offsets
+        nOffsets = 10
+        offsets = np.arange(nOffsets + 1) / nOffsets - 0.5
+        dy, dx = np.meshgrid(offsets, offsets, indexing="ij")
+        dy = dy.reshape(-1, 1, 1) # Reshape for broadcasting
+        dx = dx.reshape(-1, 1, 1)
+
+        # Distance to each subpixel
+        r = np.sqrt((x + dx) ** 2 + (y + dy) ** 2)
+
+        # Mask pixels by distance, then average number inside mask
+        inside = (r >= 0.61 * diameter / 2) & (r <= diameter / 2)
+        inside = inside.mean(axis=0).astype(float)
+
+        return inside
+
+    def correlateImage(
+        self,
+        image: np.ndarray,
+        resolution: int = 4,
+        dMin: int = 20,
+        dMax: int = 500,
+        length: int | None = None,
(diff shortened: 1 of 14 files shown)


python/lsst/ts/wep/task/donutStampSelectorTask.py
```diff
@@ -159,10 +159,8 @@ def run(self, donutStamps: DonutStamps) -> pipeBase.Struct:
             "ENTROPY",
             "FRAC_BAD_PIX",
             "MAX_POWER_GRAD",
-            "RADIUS",
-            "X_PIX_LEFT_EDGE",
-            "X_PIX_RIGHT_EDGE",
-            "RADIUS_FAIL_FLAG",
+            #            "RADIUS",
```
gmegh: why are these commented out?

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -551,6 +543,11 @@ def addVisitLevelMetadata(
         inputDonutStamps.metadata["BORESIGHT_DEC_RAD"] = (
             donutCatalog.meta["visit_info"]["boresight_dec"].to(u.rad).value
         )
+        inputDonutStamps.metadata["RADIUS"] = (
+            0.5 * donutCatalog.meta["visit_info"]["donut_diameter"]
+        )
+        # there is always a single donut diameter per detector
+        inputDonutStamps.metadata["RADIUS_FAIL_FLAG"] = np.zeros(1)
```
gmegh: well, i think we should probably remove this and fix MTAOS to not rely on RADIUS_FAIL_FLAG

In addition to that, maybe we could standardize whether we store the diameter or the radius?

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,348 @@
...
+            prom, *_ = peak_prominences(d2, peaks)
+            if len(peaks) > 0:
+                # Get first peak that is sufficiently prominent
+                sufficiently_prom = (prom / prom.max()) > 0.6
+                solution = diameters[peaks[sufficiently_prom][0]]
+
+            # If still no peaks, we have failed
+            # Note it would be very surprising to get here!
+            else:
+                solution = np.nan
```
gmegh, line 340: Maybe we could add some logging here?

python/lsst/ts/wep/task/cutOutDonutsBase.py
suberlak: ok, I can change the quantity in `donutCatalog` `visit_info`  to be also `radius`, to make it more consistent.

python/lsst/ts/wep/task/donutStampSelectorTask.py
suberlak: Good catch, I need to remove these from here (because in this place care was taken to copy over array data, whereas now `RADIUS` becomes a scalar quantity, or 1-element array so no need to loop over all donuts )

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,348 @@
...
+
+        # Normalize the image
+        image /= image[~mask].max()
+
+        return image
+
+    @staticmethod
+    def cropAndBinImage(
+        image: np.ndarray,
+        L: int | None = None,
```
jbkalmbach: Since this is a user facing function let's use a more descriptive name here.

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,348 @@
...
+
+        return inside
+
+    def correlateImage(
+        self,
+        image: np.ndarray,
+        resolution: int = 4,
+        d_min: int = 20,
+        d_max: int = 200,
+        L: int | None = None,
```
jbkalmbach: Once again, more helpful (even just `length`) variable name.

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,348 @@
...
+        if L is not None:
+            # We will crop around the brightest pixel
+
+            # Cutout pads on side so we don't select for
+            # bright donuts falling off the sensor
+            padded = image[pad:-pad, pad:-pad]
+
+            # Bin the array so we don't select individual hot pixels
+            binning_factor = 8
+            binned = binArray(padded, binning_factor)
```
jbkalmbach: `binArray` just uses a mean bin. If an individual hot pixel is much higher than the good, bright pixels this will not avoid the hot pixel. Maybe `binArray` should be updated to have mean and median options?

jbkalmbach (commented):
Each function in `DonutCorrelator` should have a test.

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: Do you have something in mind?  It does crop and bin the image after all... (i.e. it's already descriptive).

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: updated to `length` here and elsewhere

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: ok, added an option to do `median` (as well as a test for `binArray` which didn't exist)

suberlak: Shall we use `median` binning option here  by default then  ?

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: ok variable changed to `length`

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: ok, setting `median` as default

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,348 @@
...
+        # Crop and bin the image
+        image = self.cropAndBinImage(
+            image=image,
+            length=length,
+            pad=min(d_max // 2, np.min(image.shape) // 4),
+            binning=resolution,
+        )
+
+        # Start with minimum diameter
+        diam = d_min
```
gmegh: This line looks extraneous? you are overwriting the variable in the for loop later

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,348 @@
...
+
+        diameters = []
+        correlations = []
+        for diam in np.arange(d_min, d_max + resolution, resolution):
+            # Don't go past d_max
+            if diam > d_max:
```
gmegh: why do you need this if we are in a for loop?

tests/task/test_cutOutDonutsBase.py
```diff
@@ -28,6 +28,7 @@
 from lsst.daf.base import PropertySet
 from lsst.daf.butler import Butler
 from lsst.obs.lsst import LsstCam
+import lsst.geom
```
gmegh: what is this doing here?

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: because if `(d_max - d_min) % resolution` do not divide evenly, then  `np.arange` can include a final `diam` that exceeds `d_max`, and then the `if` here is useful to enforce a hard upper limit.

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: removed

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,348 @@
...
+            correlations = np.concatenate((correlations, corr[1:]))
+
+        # Find correlation peaks
+        peaks, props = find_peaks(correlations)
+        prom, *_ = peak_prominences(correlations, peaks)
+
+        # If we have peaks, select greatest prominence
+        if len(peaks) > 0 and prom.max() > 1e-2:
+            # Index of greatest prominence
+            prom, *_ = peak_prominences(correlations, peaks)
```
gmegh: why are you running this again?

tests/task/test_cutOutDonutsBase.py
suberlak: removed, seems not needed (test passes without that )

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,348 @@
...
+        diameters = np.array([], dtype=int)
+        correlations = np.array([])
+
+        # We will progress by powers of 2 in resolution
+        # and decrease resolution when resolution = 5% of diameter
+        # and max diameter we will test is longest side of image
+        L_max = max(image.shape)
+        niter = np.ceil(np.log2(0.05 * L_max)).astype(int)
+
+        for n in range(0, niter + 1):
```
gmegh: so help me understand this, why do we have a loop here and also inside the correlateImage function?

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,345 @@
...
+            # Set resolution, diameter range, cutout size
+            i_res = 2**n
```
gmegh: what is this "i" referring to? this variable name makes it hard to read, imo

suberlak: it's the i-th resolution applied, since it dynamically changes the resolution applied . In turn, resolution, d_min, d_max span a new diameter spare tested by `correlateImage`

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: Here we iterate over  donut diameter  grids  of different resolution. Then for each grid, correlateImage loops templates corresponding to probed diameters.  `correlateImage` stores the diameters and corresponding correlation values.  That's why here we  append these results of different grids : ```     # Append to our existing arrays
            diameters = np.concatenate((diameters, diam[1:]))
            correlations = np.concatenate((correlations, corr[1:]))```

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: because it's needed to get the greatest prominence

gmegh: look at line 315 you are running the same line

python/lsst/ts/wep/donutSizeCorrelator.py
gmegh: I think "i_min" is not very clear then, maybe `i_dim_min`?

suberlak: updated variable names

gmegh:
LGTM. Please squash commits before merging

gmegh (approved):

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,344 @@
...
+        """Prep image from the butler for correlation.
+
+        Parameters
+        ----------
+        exposure : lsst.afw.image.Exposure
+            Exposure from the butler.
+        badPixelTypes : tuple[str], optional
+            Names of flags that will be masked.
+            Default is ("SAT", "BAD", "NO_DATA", "INTRP")
+        nDilation : int, optional
```
jbkalmbach, line 50: Why such a large number of dilation iterations for a bad pixel?

python/lsst/ts/wep/donutSizeCorrelator.py
```diff
@@ -0,0 +1,344 @@
...
+        np.ndarray
+            Fractional donut mask
+        """
+        # Create grid of pixel centers
+        x = np.arange(diameter + 5, dtype=float)
+        x -= x.mean()
+        x, y = np.meshgrid(x, x)
+
+        # Build sub-pixel offsets
+        N_offsets = 10
```
jbkalmbach: The variable names in this class are a mix of snake case and camel case. It needs to be consistent. In `ts_wep` we have used camel case in our other classes.

jbkalmbach: See also `d_min`, `d_max` below

python/lsst/ts/wep/donutSizeCorrelator.py
suberlak: Operating on an entire image this removes a region surrounding it, it doesn't actually end up being a large footprint

tests/test_donutSizeCorrelator.py
```diff
@@ -0,0 +1,163 @@
...
+        opticalModel='offAxis',
+        padding=5,
+        isBinary=True,
+        )
+        diameter, diameters, correlations = self.correlator.getDonutDiameter(template)
+
+        self.assertTrue(np.isfinite(diameter))
+        self.assertIsInstance(diameter, (int, float, np.integer))
+        self.assertGreater(diameter, 0)
+        self.assertGreater(diameter, 200)
```
jbkalmbach, line 155: Why do you have two `assertGreater`? If it passes the second it would necessarily pass the first wouldn't it?

suberlak: ok, removed the first assertion

PR #25, opened 2025-09-09
Fix CalcZernikesUnpairedTask to handle numCores parameter in estimate…

jbkalmbach:
…Zernikes.


python/lsst/ts/wep/task/calcZernikesUnpairedTask.py
```diff
@@ -109,26 +148,35 @@ def run(self, donutStamps: DonutStamps) -> pipeBase.Struct:
             donutQualityTable = QTable([])
 
         # Assign stamps to either intra or extra
-        if selectedDonuts[0].wep_im.defocalType == DefocalType.Extra:
+
```
gmegh: remove extra line?

python/lsst/ts/wep/task/calcZernikesUnpairedTask.py
```diff
@@ -109,26 +148,35 @@ def run(self, donutStamps: DonutStamps) -> pipeBase.Struct:
...
+        defocalType = donutStamps.metadata["DFC_TYPE"]
+        if defocalType == "extra":
             self.stampsExtra = selectedDonuts
-            extraStamps = selectedDonuts
-            intraStamps = DonutStamps([])
+            self.stampsIntra = None
```
gmegh: why are you defining this as None at all? It doesn't seem to be used

tests/testData/gen3TestRepo/gen3.sqlite3
gmegh, line 1: why is this being changed at all?

tests/task/test_cutOutDonutsUnpairedTask.py
```diff
@@ -171,9 +171,14 @@ def testPipeline(self) -> None:
         ) = self._getDataFromButler()
 
         # Test normal behavior
-        taskOut = self.task.run(
-            [copy(exposureExtra), copy(exposureIntra)],
-            [donutCatalogExtra, donutCatalogIntra],
+        taskOutExtra = self.task.run(
+            copy(exposureExtra),
+            donutCatalogExtra,
```
gmegh: do you also need to deepcopy the catalog?

python/lsst/ts/wep/task/calcZernikesUnpairedTask.py
jbkalmbach: It's used when we run `createUnpairedZernikeTable` below.

tests/testData/gen3TestRepo/gen3.sqlite3
jbkalmbach: Had to redefine the dimensions for unpaired cutouts

tests/task/test_cutOutDonutsUnpairedTask.py
jbkalmbach: Not here. I'm not testing the catalog at all, just the cutouts.

gmegh (approved):
LGTM, but you are still changing the gen3 file, which I don't think you should be changing

python/lsst/ts/wep/task/calcZernikesUnpairedTask.py
```diff
@@ -84,11 +83,52 @@ class CalcZernikesUnpairedTask(CalcZernikesTask):
...
+        zkCoeffCombined
+            Combined zernikes returned by self.combineZernikes.run(...)
+
+        Returns
+        -------
+        table : `astropy.table.QTable`
+            Table with the Zernike coefficients
+        """
+
+        if self.stampsExtra is None:
```
gmegh: Actually, one second, why don't you just pass self.stampsIntra and self.stampsExtra directly to the function, and then in the run method populate them with DonutStamps([]) instead of None

jbkalmbach: Ok, thanks. I was annoyed by this and never came back to it. I think I've come up with a cleaner solution now.

python/lsst/ts/wep/task/calcZernikesUnpairedTask.py
```diff
@@ -85,10 +84,20 @@ class CalcZernikesUnpairedTask(CalcZernikesTask):
...
-        # If no donuts are in the donutCatalog for a set of exposures
-        # then return the Zernike coefficients as nan.
+    def run(
+        self,
+        donutStamps: DonutStamps,
+        numCores: int = 1,
+    ) -> pipeBase.Struct:
+        # Assign stamps to either intra or extra
+        defocalType = donutStamps.metadata["DFC_TYPE"]
+        if defocalType == "extra":
```
gmegh: Sorry, one last comment, you probably don't need this here, since you are doing it right after the donutStampSelector again,no?

jbkalmbach: After `donutStampSelector` I change it to only include the selected donuts.

gmegh: Right, so why don't you just assign it after running donutStampSelector? My point is that we probably don't need this assignment here and can just use the one you are doing again below after the donutStampSelector

PR #26, opened 2025-09-11
Tickets/DM-52459: add donut_id

suberlak:
Adding `donut_id`  based on detector and donut brightness.

### DIFF
--- a/python/lsst/ts/wep/task/donutStamp.py
+++ b/python/lsst/ts/wep/task/donutStamp.py
@@ -74,6 +74,8 @@ class DonutStamp(AbstractStamp):
         are available camera names currently.
     bandpass : `str`
         The bandpass for the stamp image.
+    donut_id: `str`
+        The donut id defined at detection stage.
     archive_element : `afwTable.io.Persistable`, optional
         Archive element (e.g. Transform or WCS) associated with this stamp.
         (the default is None.)
@@ -83,6 +85,7 @@ class DonutStamp(AbstractStamp):
         camera coordinate system (CCS), with the CWFSs rotated to the same
         orientation as the science sensors. It is this object that will be used
         to interface with the wavefront estimator.
+
     """
 
     stamp_im: afwImage.MaskedImageF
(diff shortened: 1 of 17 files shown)


python/lsst/ts/wep/task/donutStamp.py
```diff
@@ -82,6 +84,7 @@ class DonutStamp(AbstractStamp):
         camera coordinate system (CCS), with the CWFSs rotated to the same
         orientation as the science sensors. It is this object that will be used
         to interface with the wavefront estimator.
+
```
jbkalmbach, line 88: Remove blank line.

tests/task/test_donutStamp.py
```diff
@@ -121,6 +123,9 @@ def testFactory(self) -> None:
             bandpass = donutStamp.bandpass
             self.assertEqual(bandpass, "r")
 
+            donutId = donutStamp.donut_id
```
jbkalmbach, line 127: I think you could remove the blank line above this since it goes with testing the properties.

jbkalmbach (commented):
Why did you decide to make this a string instead of an `int`?

jbkalmbach (commented):
Do we care that the same donut could have different `donut_id` values in an extra-intra pair when running in full array mode? Probably not I think but can you think of any reason?

suberlak:
I went along with a `str` because it allowed a possibility of making names like eg. `191_0001` or `R00_191_0001` etc.  But I could revert to int if you think it's better. It is possible to fold in a visit number (since I have access to `exposure` in [generateDonutCatalogUtils.py](https://github.com/lsst-ts/ts_wep/pull/26/files#diff-92796eb81d5b5a5575f214848aec0ac1236152a3c684af0ba45e571a166a3ac9)), i.e. make it not only `2025053100117_191_0001`, but `202505310011701910001`  (eg. 4 spaces for detector and 4 spaces for donut for clearer reading)- that way we'd have visit-detector-donut granularity (i.e. possible to distinguish intra/extra when running  in FAM mode since extra/intra have different visit number).

jbkalmbach (commented):
It looks like you need to add a test somewhere that verifies donut_id is output in the format you expect. I don't see a test for that anywhere.

suberlak:
added test for donut_id

jbkalmbach (approved):
Thanks, looks good.

PR #27, opened 2025-09-15
Fix missing logic in generateDonutFromRefitWcsTask that caused failur…

jbkalmbach:
…es when scatter in WCS fit was too large.


jfcrenshaw (approved):
LGTM

PR #28, opened 2025-09-17
Merge AiDonut model

jfcrenshaw:
Includes an estimation module and an estimate Zernikes subtask.

Note the testing isn't that extensive at the moment. I think we can write more tests once we figure out a way to properly store the model file (or a small test version) so it can be used in CI. However, per discussions with Merlin and Guillem, we are deciding to merge the classes we have so there is more stability in the pipeline infrastructure visible from the summit.

Note the new pipetask has been tested and validated on BTS.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,15 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-14.17.0:
+
+-------------
+ 14.17.0
+-------------
+
+* Add AiDonutAlgorithm for wavefront estimation using a PyTorch model.
+* Add EstimateZernikesAiDonutTask to use AiDonutAlgorithm.
+
 .. _lsst.ts.wep-14.16.1:
 
 -------------
--- a/python/lsst/ts/wep/estimation/aiDonut.py
+++ b/python/lsst/ts/wep/estimation/aiDonut.py
@@ -1,0 +1,245 @@
+# This file is part of ts_wep.
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import numpy as np
+import torch
+from lsst.ts.wep import Image, Instrument
+from lsst.ts.wep.estimation.wfAlgorithm import WfAlgorithm
+from lsst.ts.wep.utils import getModulePath, makeDense, makeSparse
+
+__all__ = ["AiDonutAlgorithm"]
+
+DEFAULT_MODEL_PATH = (
+    getModulePath() + "/tests/testData/testAiModels/test_aidonut_model_file.pt"
+)
+
+
+class AiDonutAlgorithm(WfAlgorithm):
+    """Wavefront estimation using a PyTorch model.
+
+    Parameters
+    ----------
+    modelPath : str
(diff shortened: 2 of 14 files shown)


doc/versionHistory.rst
```diff
@@ -5,11 +5,22 @@
 ##################
 Version History
 ##################
+
+.. _lsst.ts.wep-14.17.0:
```
gmegh, line 9: I wonder if this should be 15.0? It's an important update to make it work with AI, so just wondering, and we've had a lot of updates since the last major release?

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -0,0 +1,239 @@
...
+    - must have a `nollIndices` attribute that lists the Noll indices
+    corresponding to the output coefficients.
+    """
+
+    def __init__(
+        self,
+        modelPath: str,
+        device: str = "cpu",
+    ) -> None:
+        # Save parameters
```
gmegh: maybe we can remove this comment?

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -0,0 +1,239 @@
...
+        Parameters
+        ----------
+        value : str
+            Path to the PyTorch model file.
+        """
+        # Load the model
+        self.model = torch.load(value, map_location=self.device, weights_only=False)
+        self.model.eval()  # Put in evaluation mode
+
+        # If loading the model succeeded, save the path
```
gmegh, line 114: is there an option it doesn't succeed? If so, should you raise and error if that's the case?

tests/estimation/test_aidonut.py
```diff
@@ -0,0 +1,41 @@
...
+
+import unittest
+
+from lsst.ts.wep.utils.testUtils import enforce_single_threading
+
+enforce_single_threading()
+
+from lsst.ts.wep.estimation import AiDonutAlgorithm  # noqa: E402
+
+class TestAiDonutAlgorithm(unittest.TestCase):
```
gmegh, line 32: I would suggest probably adding another test that uses some random weights and makes sure the outputs and inputs are working fine?

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -0,0 +1,239 @@
...
+                fys_tch,
+                focalFlags_tch,
+                bands_tch,
+            )
+        outputs = outputs.cpu().numpy()  # shape: (2, n_zernikes)
+
+        # Average the two estimates
+        zk = outputs.mean(axis=0)
+
+        # Only return requested nollIndices
```
gmegh, line 223: Given the if statement you have at the beginning of this method, why are these two lines needed?

gmegh (commented):
Looking good, just left some comments

doc/versionHistory.rst
jfcrenshaw: Typically major release numbers correspond to backwards incompatible changes, and minor release numbers correspond to new features, so I think this is the right version number.

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: That already happens -- if `torch.load()` fails, it will raise an error and tell you why it failed

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: The if statement only checks that the requested Noll indices are a subset of the allowed Noll indices. These lines actually take care of retrieving the requested Noll indices

tests/estimation/test_aidonut.py
jfcrenshaw: That's a non-trivial thing to do. For now the plan is to merge with the understanding it has been demonstrated to work on local tests, BTS, and the summit, and when we make a final decision about how to store model files, we will add a small dummy test file that we can write more extensive tests around.

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: Done.

tests/estimation/test_aidonut.py
gmegh: It seems to me that you are referring to two different things. Wouldn't the dummy test file be stored in testData. I would have expected it should just have the same architecture but filled with some random values, so it should not occupy that much space. 

Maybe by non-trivial, you were referring to how to create the dummy file? I would expect the tests to be fairly similar to the ones of danish estimator. To begin with maybe just running the task with this algorithm and checking the metadata would be enough

gmegh (approved):
LGTM! Thanks for making that change

PR #34, opened 2025-10-29
Tickets/DM-53082: fix donut mask in calculateSN

suberlak:
This fixes donut mask selection in cutOutDonutsBase

### DIFF
--- a/python/lsst/ts/wep/task/cutOutDonutsBase.py
+++ b/python/lsst/ts/wep/task/cutOutDonutsBase.py
@@ -341,21 +341,46 @@ class CutOutDonutsBaseTask(pipeBase.PipelineTask):
         mask = stamp.stamp_im.mask
         varianceArray = stamp.stamp_im.variance.array
 
-        # The following are example donut mask values:
-        # maskPlaneDict={'BAD': 0, 'BLEND': 10, 'CR': 3, 'DETECTED': 5,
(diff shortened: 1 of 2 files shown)


jbkalmbach (approved):
Looks ok to me.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -341,20 +341,40 @@ def calculateSN(self, stamp: DonutStamp) -> dict:
...
+        #      10000000000000000  (65536 - only PARTLY_VIGNETTED)
+        #    & 00010000000000000  (8192 - DONUT bit mask)
+        #    -------------------
+        #      00000000000000000  (0 - result is zero)
+
+        donutPlaneBit = mask.getPlaneBitMask("DONUT")
+        blendBit = mask.getPlaneBitMask("BLEND")
+        # Select pixels with DONUT set AND BLEND not set
+        hasDonut = (mask.array & donutPlaneBit) > 0
+        notBlend = (mask.array & blendBit) == 0
```
jbkalmbach, line 379: Do you want to consider also avoiding `SAT`, `CR`, etc. here as well?

suberlak: Sure we can do that - just to be sure  I tested the impact on SN calculation. Here's a comparison for 2625 donuts from 10/23 for which previously we had SN as NaN.  This would affect the signal portion.  First we avoid only BLEND (blue), then avoid BLEND and CR and SAT (orange). 
unknown

PR #35, opened 2025-10-31
Added height map to instrument

jfcrenshaw:
Here is the interpolated map:
image

You can confirm this map is in the CCS because of the visible e2v raft sticking out in the bottom right.

Compare this to the map provided by the camera team (notice this one is transposed from CCS):
image

Here is a pair of donuts simulated with a constant focal plane height offset of +0.5mm. That positive offset moves the sensors away from the camera, i.e. towards the M1M3. That means the extrafocal donut is 0.5mm closer to focus, whereas the intrafocal donut is an additional 0.5mm farther from focus. You can see this borne out in the simulated images:
image

### DIFF
--- a/python/lsst/ts/wep/instrument.py
+++ b/python/lsst/ts/wep/instrument.py
@@ -29,8 +29,9 @@ from typing import Sequence
 
 import batoid
 import numpy as np
+from astropy.table import Table
 from lsst.ts.wep.utils.enumUtils import BandLabel, DefocalType, EnumDict
-from lsst.ts.wep.utils.ioUtils import mergeConfigWithFile
+from lsst.ts.wep.utils.ioUtils import mergeConfigWithFile, resolveRelativeConfigPath
 from scipy.optimize import minimize_scalar
 
 from .utils.zernikeUtils import createZernikeBasis, createZernikeGradBasis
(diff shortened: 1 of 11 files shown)


python/lsst/ts/wep/instrument.py
```diff
@@ -104,6 +105,11 @@ class Instrument:
         detector offset is then used for everything else. Note that
         depending on the model, the sign of this value might matter.
         (the default is None)
+    heightMap : str or astropy.table.Table or None, optional
```
jmeyers314, line 108: Somewhere we should write down the units in addition to the dimensions.  (mm, micron, ?).
Also the coordinate system.

jmeyers314 (approved):
Looks good to me with caveats that we discussed on slack (.e.g, I think this should really be an intrinsic zernikes calibration product).

I can't tell at a glance; did you consider whether any of this can/should be cached?

python/lsst/ts/wep/instrument.py
jfcrenshaw: The units are handeled explicitly below. As long as the values have some astropy unit of length attached, it will handle the rest.
https://github.com/lsst-ts/ts_wep/blob/99f0c066f9d89b6b3389c0990f4ecf37603f8e32/python/lsst/ts/wep/instrument.py#L747-L749

I'll make a comment about the coordinate system.

jfcrenshaw:
Re: the question about it being cached, I believe it essentially is. The height map is only constructed once when loading the instrument, which takes ~100ms in a USDF notebook.

Then the height map is called inside `getOffAxisCoeff`, which is cached.

jfcrenshaw:
Fixed sign error. +z in height map corresponds to extrafocal in the camera data. It now also corresponds to extrafocal in the WEP instrument too.

Here for a +0.5mm offset you can see the extrafocal donut is larger as the extrafocal sensor has been pushed farther from focus.
image

jfcrenshaw:
Implemented at the CalcZernikesTask level instead

PR #36, opened 2025-11-02
Correct FAM by shifting whole camera

jfcrenshaw:
Let's make sure our plots of the difference in intrinsics match.

Here I am plotting
- extrafocal
- LSSTCam shift - Detector shift
- x and y axes are field angles

image

This is a pretty close match to the plot you shared on slack (copied here), where you stated all higher-order residuals are close to zero:
image

The first thing to notice is our z4 residual has opposite signs. I assume you plotted Detector shift - LSSTCam shift? The other thing is we don't have the same parities for z5 and z8. This is probably related to coordinate systems?

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -19,10 +19,16 @@ def pytest_addoption(parser: Parser) -> None:
         default=False,
         help="Run WEP pipeline before all tests.",
     )
+    parser.addoption(
+        "--skip-pretest",
+        action="store_true",
+        default=False,
+        help="Skip pre-test WEP pipeline run.",
+    )
 
 
 def pytest_configure(config: Config) -> None:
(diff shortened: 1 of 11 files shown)


jmeyers314:
LGTM.  I'll think about the coordinate systems and get back to you this afternoon.

jmeyers314:
> The first thing to notice is our z4 residual has opposite signs. I assume you plotted Detector shift - LSSTCam shift? 

Yes.

> The other thing is we don't have the same parities for z5 and z8. This is probably related to coordinate systems?

I tried (briefly) reproducing yours and couldn't, but I assume this is correct.

jmeyers314 (approved):
LGTM.  Thanks!

PR #37, opened 2025-11-06
DM-53183: Upgrade WEP tasks to return OPD, intrinsics, and deviation

jbkalmbach:


jfcrenshaw:
I put in a dummy implementation that I think will work. Tomorrow I will work on getting real files loaded, and writing/fixing tests to make sure I did everything correctly.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -262,54 +369,29 @@ def createZkTable(
...
+
+            # Unpack data from stamps, handling cases with None stamps
+            intraAngle, intraCentroid, intraIntrinsics = self._unpackStampData(intra)
+            extraAngle, extraCentroid, extraIntrinsics = self._unpackStampData(extra)
+
+            # Average the intrinsics
+            intrinsics = np.nanmean((intraIntrinsics, extraIntrinsics), axis=0) * u.micron
+
+            # Calculate the wavefront deviation
+            deviation = zk - intrinsics
```
jbkalmbach, line 383: What's the motivation behind carrying around the deviation if it's easily derived when needed?

gmegh: I think it's for compatibility with what we were doing previously and because it doesn't hurt to have it stored in the columns

jfcrenshaw: I agree, it's minimal additional storage, makes things more transparent, and simplifies downstream code.

python/lsst/ts/wep/task/combineZernikesMeanTask.py
```diff
@@ -31,5 +32,13 @@ class CombineZernikesMeanTask(CombineZernikesBaseTask):
...
 
-    def combineZernikes(self, zernikeArray: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
-        return np.mean(zernikeArray, axis=0), np.zeros(len(zernikeArray), dtype=int)
+    def _combineZernikes(self, zkTable: Table) -> Table:
+        # Set all donuts to used (no rejection in this simple mean)
+        zkTable["used"] = True
+
+        # Calculate mean of every Zernike column
+        for colName in zkTable.colnames:
+            if colName.startswith("Z"):
```
jmeyers314: Are we ready to use `zkTable.meta["nollIndices"]` here instead of `startswith` filtering?

python/lsst/ts/wep/task/combineZernikesSigmaClipTask.py
```diff
@@ -73,26 +81,42 @@ def __init__(self, **kwargs: Any) -> None:
             self.sigmaClipKwargs[key] = val
         self.maxZernClip = self.config.maxZernClip
         self.stdMin = self.config.stdMin
+        self.zkClipType = self.config.zkClipType
+
+    def _combineZernikes(self, zkTable: Table) -> Table:
+        # Extract the array of Zernikes to perform clipping on
+        if self.zkClipType == "opd":
```
jmeyers314, line 87: Prefer to use `zkTable.meta["nollIndices"]` to regexes if possible.

python/lsst/ts/wep/task/combineZernikesBase.py
```diff
@@ -48,73 +49,100 @@ def __init__(self, **kwargs: Any) -> None:
         pipeBase.Task.__init__(self, **kwargs)
         self.log = logging.getLogger(type(self).__name__)  # type: ignore
 
-    def run(self, zernikeArray: np.ndarray) -> pipeBase.Struct:
+    @staticmethod
+    def _setAvg(zkTable: Table, colName: str, function: Callable, useIdx: list | None = None) -> None:
```
jmeyers314: Can we make the function default `numpy.nanmean`?

jmeyers314 (approved):
LGTM assuming all the tests pass.  I left a few small comments.

python/lsst/ts/wep/task/combineZernikesMeanTask.py
jfcrenshaw: Yeah I wrote that as a catch-all for all Zernike-type columns. But It's probably better to explicitly loop over all of those columns.

python/lsst/ts/wep/task/combineZernikesSigmaClipTask.py
jfcrenshaw: Ah yes, nice catch. I forgot to go back and replace these!

python/lsst/ts/wep/task/combineZernikesBase.py
jfcrenshaw: Done.

PR #38, opened 2025-11-18
DM-53322 Tarts step1a error

PetchMa:
… to put back the average row and comply with parent class method signature

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -13,6 +13,11 @@ Version History
 -------------
 
 * To ensure bakcwards compatibility with DM stack < w_2025_43 remove the change in 16.0.1 update for centroid error limit and only add it in test pipeline as an additional config.
+* Fix CalcZernikesNeuralTask to handle missing intrinsic maps by overriding
+  `_unpackStampData()` method to return NaN for intrinsic values (neural task
+  does not use intrinsic Zernike tables).
+* Fix CalcZernikesNeuralTask `createZkTable()` call to remove unsupported
+  `zkCoeffCombined` argument that was causing TypeError in production pipelines.
 
 .. _lsst.ts.wep-16.1.0:
 
@@ -38,11 +43,6 @@ Version History
 -------------
 
 * Update astromTask defaults inside refitWcsTask to turn off new default centroid error limit added in DM stack w_2025_46.
-* Fix CalcZernikesNeuralTask to handle missing intrinsic maps by overriding
-  `_unpackStampData()` method to return NaN for intrinsic values (neural task
-  does not use intrinsic Zernike tables).
-* Fix CalcZernikesNeuralTask `createZkTable()` call to remove unsupported
-  `zkCoeffCombined` argument that was causing TypeError in production pipelines.
 
 .. _lsst.ts.wep-16.0.0:
 
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -38,6 +38,11 @@ Version History
 -------------
 
 * Update astromTask defaults inside refitWcsTask to turn off new default centroid error limit added in DM stack w_2025_46.
+* Fix CalcZernikesNeuralTask to handle missing intrinsic maps by overriding
+  `_unpackStampData()` method to return NaN for intrinsic values (neural task
+  does not use intrinsic Zernike tables).
+* Fix CalcZernikesNeuralTask `createZkTable()` call to remove unsupported
+  `zkCoeffCombined` argument that was causing TypeError in production pipelines.
 
 .. _lsst.ts.wep-16.0.0:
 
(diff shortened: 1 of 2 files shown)


PetchMa:
@mfisherlevine I just tested on USDF and produced plots like before, could I get a review please? and perhaps a redeployment to BTS? (with TARTS most updated version?)

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
```diff
@@ -1251,3 +1244,74 @@ def run(
...
+                f"{len(agg)} does not match configured nollIndices length {len(self.nollIndices)}."
+            )
+
+        agg_quant = (agg * u.micron).to(u.nm)
+        avg_row = zkTable[0]
+
+        for idx, j in enumerate(self.nollIndices):
+            avg_row[f"Z{j}"] = agg_quant[idx]
+            avg_row[f"Z{j}_deviation"] = agg_quant[idx]
+            avg_row[f"Z{j}_intrinsic"] = np.nan * u.nm
```
gmegh: I would say you should use what calcZernikes is doing to populate this?

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
```diff
@@ -1251,3 +1244,74 @@ def run(
...
+            Table returned by ``createZkTable`` containing placeholder values.
+            The first row corresponds to the aggregate (average) entry.
+        aggregatedZernikes : `numpy.ndarray`
+            One-dimensional array of aggregated Zernike coefficients produced
+            by TARTS (in microns).
+        """
+        if len(zkTable) == 0:
+            return
+
+        agg = np.asarray(aggregatedZernikes, dtype=float).ravel()
```
gmegh: isn't aggregatedZernikes already an array?

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
```diff
@@ -1251,3 +1244,74 @@ def run(
             donutTable=donutTable,
             donutQualityTable=donutQualityTable,
         )
+
+    def _unpackStampData(self, stamp: DonutStamp | None) -> tuple:
```
gmegh, line 1248: you are not using this function anywhere....

gmegh (changes requested):
left some comments, you define a function that you don't use anywhere

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
PetchMa: Not sure this is dead code? _unpackStampData() is indirectly called from the parent CalcZernikesTask.createZkTable() (see python/lsst/ts/wep/task/calcZernikesTask.py, L356‑405). Because CalcZernikesNeuralTask reuses the parent’s table builder, overriding _unpackStampData() is the only way to avoid touching self.intrinsicMapIntra/Extra (which the neural implementation never populates) and still satisfy the base class’s expectations. without this override, createZkTable() will try to look up attributes that don’t exist and crash (which is what you saw on the on‑sky run). does this make sense?

doc/versionHistory.rst
```diff
@@ -30,6 +30,11 @@ Version History
 -------------
 
 * Update astromTask defaults inside refitWcsTask to turn off new default centroid error limit added in DM stack w_2025_46.
+* Fix CalcZernikesNeuralTask to handle missing intrinsic maps by overriding
```
gmegh: We are currently at version 16.1 and you are adding this to 16.0.1? That doesn’t make sense?

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
```diff
@@ -1251,3 +1244,80 @@ def run(
...
+
+        agg_quant = (agg * u.micron).to(u.nm)
+        avg_row = zkTable[0]
+
+        for value, column in zip(agg_quant, opd_columns):
+            avg_row[column] = value
+        for column in intrinsic_columns:
+            avg_row[column] = np.nan * u.nm
+        for value, column in zip(agg_quant, deviation_columns):
+            avg_row[column] = value
```
gmegh: Upon further reflection, i think you should only be populating opdAcolumns and leaving the rest empty for now

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
```diff
@@ -1251,3 +1244,80 @@ def run(
...
+        agg = aggregatedZernikes[0]
+        if len(agg) != len(opd_columns):
+            raise ValueError(
+                "Neural aggregated Zernike vector length "
+                f"{len(agg)} does not match expected "
+                f"{len(opd_columns)} OPD columns."
+            )
+
+        agg_quant = (agg * u.micron).to(u.nm)
+        avg_row = zkTable[0]
```
gmegh: It'd be safer to query the row that has label 'average' no?

gmegh (approved):
One last comment. Please address it before merging. Also make sure to squash your commits, please before merging

PR #39, opened 2025-11-18
Update astromTask defaults inside refitWcsTask to turn off new defaul…

jbkalmbach:
…t centroid error limit added in DM stack w_2025_46.


suberlak (approved):
LGTM!

PR #49, opened 2025-12-09
Make reassignCwfsCutoutsTask reassign cutouts to both intra- and extr…

gmegh:
…a-focal CWFS detectors


jbkalmbach:
@gmegh Should we close this? I think all these changes are in `develop` now from the `danish_struts` merge?

PR #52, opened 2025-12-16
Fix bug with intrinsic Zernikes in unpaired calc Zernikes task

jfcrenshaw:
The problem was with `np.full_like(self.nollIndices, np.nan)`. self.nollIndices was filled with integers, so it tried to cast np.nan to an integer, which results in bad, unpredictable behavior.

I wrote a test to catch this -- specifically by setting the Z4 intrinsic to a known value, making sure it was reproduced in the Zernike table. Then I implemented the fix so the test would pass.

### DIFF
--- a/python/lsst/ts/wep/task/calcZernikesTask.py
+++ b/python/lsst/ts/wep/task/calcZernikesTask.py
@@ -250,7 +250,7 @@ class CalcZernikesTask(pipeBase.PipelineTask, metaclass=abc.ABCMeta):
         if stamp is None:
             fieldAngle = np.array(np.nan, dtype=pos2f_dtype) * u.deg
             centroid = np.array((np.nan, np.nan), dtype=pos2f_dtype) * u.pixel
-            intrinsics = np.full_like(self.nollIndices, np.nan) * u.micron
+            intrinsics = np.full(len(self.nollIndices), np.nan) * u.micron
         else:
             fieldAngle = np.array(stamp.calcFieldXY(), dtype=pos2f_dtype) * u.deg
             centroid = (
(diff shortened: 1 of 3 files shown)


python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -250,7 +250,7 @@ def _unpackStampData(self, stamp: DonutStamp) -> tuple[u.Quantity, u.Quantity, u
         if stamp is None:
             fieldAngle = np.array(np.nan, dtype=pos2f_dtype) * u.deg
             centroid = np.array((np.nan, np.nan), dtype=pos2f_dtype) * u.pixel
-            intrinsics = np.full_like(self.nollIndices, np.nan) * u.micron
+            intrinsics = np.full(len(self.nollIndices), np.nan) * u.micron
```
suberlak, line 253: important change: if `self.nollIndices` has an integer dtype (which it likely does), then np.nan cannot be represented. This means that numpy would try to cast np.nan to the integer dtype, which would result in 0 or an invalid value. Eg. 
```
self.nollIndices = np.array([4, 5, 6], dtype=int)
np.full_like(self.nollIndices, np.nan)
```
would previously yield

``` array([0, 0, 0]) ```

**Review by suberlak — APPROVED** [review]
LGTM! Thanks for catching this subtle bug!

File: python/lsst/ts/wep/task/calcZernikesTask.py (hunk 1 above)
Comment (suberlak) [reply python/lsst/ts/wep/task/calcZernikesTask.py:253]: Making it 
```
intrinsics = np.full(self.nollIndices.shape, np.nan, dtype=int) * u.micron
```
is even more explicit

PR #65, opened 2026-01-29
Tickets/dm 53904

jbkalmbach:


python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
```diff
@@ -168,9 +187,16 @@ def run(
...
         detector = exposure.getDetector()
         detectorWcs = exposure.getWcs()
-        anyFilterMapsToThis = self.config.anyFilterMapsToThis
         edgeMargin = self.config.edgeMargin
-        filterName = exposure.filter.bandLabel if anyFilterMapsToThis is None else anyFilterMapsToThis
+        filterName = exposure.filter.bandLabel
+        filterList = self.config.catalogFilterList
+        # Check that specified filter exists in catalogs
+        if self.config.photoRefFilter is not None and self.config.photoRefFilterPrefix is not None:
+            raise ValueError("photoRefFilter and photoRefFilterConfig cannot both be set.")
```
suberlak: I think `config` is a typo - did you mean `photoRefFilterPrefix` ?

python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
```diff
@@ -196,7 +239,17 @@ def run(
             blendCentersX = None
             blendCentersY = None
 
-        fieldObjects = donutCatalogToAstropy(refSelection, filterName, blendCentersX, blendCentersY)
+            # Record failure in task metadata
```
suberlak, line 242: we're catching this `RuntimeError`, what about `ValueError` - when both photoRefFilter and photoRefFilterPrefix are set (line 195) and 
`TaskError` - when requested filters don't exist in the catalog (line 222)? Shouldn't we also set metadata then?

suberlak: Eg. 
```try:
    #  ...
    self.metadata["refCatalogsPresent"] = True
except (RuntimeError, ValueError, TaskError) as e:
    self.logger.warning(f"Error processing reference catalog: {e}")
    self.metadata["refCatalogsPresent"] = False```

suberlak (approved):
Looks good, few small comments!

python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
jbkalmbach: In those cases the error is raised before we get to a point where we look for reference catalogs so we never make it this far in the code.

jbkalmbach: The user when they see that error was raised can easily find the failure and no catalogs are produced so it's not necessary to record what kind of catalog it is in the task metadata.

python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
jbkalmbach: Good catch, thanks.

PR #66, opened 2026-02-05
Tickets/DM-54058: make intrinsicTables optional

suberlak:


python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -76,6 +76,7 @@ class CalcZernikesTaskConnections(
         storageClass="ArrowAstropy",
         name="intrinsic_aberrations_temp",
         multiple=True,
```
jbkalmbach, line 78: I don't think you should touch these at all to fix LATISS. You showed me that latiss only needs to run inline via the `run` method not through pipe tasks.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -119,7 +120,8 @@ def adjust_all_quanta(self, adjuster: pipeBase.QuantaAdjuster) -> None:
                 )
 
                 intra_inputs = adjuster.get_inputs(intra_focal_data_id)
```
jbkalmbach, line 122: I think you should remove this change for the same reasoning as above.

jbkalmbach (changes requested):
Why do we have to do these changes here? Why can't you just submit a PR in `latiss_wep_align` for arrays of zeros as the intrinsic tables?

PR #68, opened 2026-02-13
Tickets/DM-53585: Add blur clipping and return chi2 in danish

jbkalmbach:


python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -517,6 +523,35 @@ def empty(
...
+        -------
+        astropy.table.QTable
+            Zernike table where donuts with outlier donut blur values
+            have been changed to false and the average recomputed.
+        """
+
+        useIdx = np.where((zkTable["used"]) & (zkTable["label"] != "average"))[0]
+        # account for average row with "- 1" on index below
+        fwhmList = np.array(zkTable.meta["estimatorInfo"]["fwhm"])[useIdx - 1]
+        blurMask = sigma_clip(fwhmList, stdfunc="mad_std", sigma_lower=99).mask
```
suberlak, line 546: seems like an aggressive threshold, could you explain how it was chosen ?

tests/task/test_combineZernikesMeanTask.py
```diff
@@ -34,7 +34,7 @@ def setUp(self) -> None:
 
     def prepareTestTable(self) -> Table:
         label = ["average"] + [f"pair{i}" for i in range(10)]
-        used = [True] + 10 * [False]
+        used = [True] + 10 * [True]
```
suberlak, line 37: why was that change needed ?

suberlak (approved):
Once chi2-like quantity  is documented we can merge. I changed PR title to be more descriptive.

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: It's actually a very loose threshold on the lower bound (will only clip if 99 times the standard deviation lower so basically never). Focused on clipping the upper values because they indicate that danish is modeling additional blur to compensate for some strange behavior in the donut stamp.

tests/task/test_combineZernikesMeanTask.py
jbkalmbach: `combineZernikesMeanTask` no longer automatically sets all donuts to `used` and instead uses what is passed from the catalog.

PR #69, opened 2026-02-17
Tickets/DM-54173: fix refitWcsTask

suberlak:
Adding clipping sources to detector boundary, so that even if donut selection is off (eg. to return as many sources as possible), the resulting catalog is still within the bounding box. Snapshot of  a post_isr_image with before / after changes, and reproducible pipetask call,  in the ticket comments https://rubinobs.atlassian.net/browse/DM-54173

### DIFF
--- a/python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
+++ b/python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
@@ -300,6 +300,50 @@ class GenerateDonutFromRefitWcsTask(GenerateDonutCatalogWcsTask):
 
         return sourceCat
 
+    def _clipDonutCatalogToDetector(
+        self,
+        donutCatalog: QTable,
+        detector: lsst.afw.cameraGeom.Detector,
+    ) -> QTable:
+        """Clip donut catalog to only include sources within the
+        detector bounding box, eroded by the configured edgeMargin.
+
+        This acts as a final safety net to ensure no donuts outside
+        the detector bounds make it into the output catalog,
+        regardless of which code path produced the catalog.
+
+        Parameters
+        ----------
+        donutCatalog : `astropy.table.QTable`
+            Donut catalog to clip.
+        detector : `lsst.afw.cameraGeom.Detector`
+            Detector object from the camera.
+
+        Returns
+        -------
+        clippedCatalog : `astropy.table.QTable`
+            Catalog with only sources inside the eroded bounding box.
+        """
+        bbox = detector.getBBox()
+        erodedBBox = bbox.erodedBy(self.config.edgeMargin)
+
+        inBounds = np.array(
+            [
+                erodedBBox.contains(lsst.geom.Point2I(int(x), int(y)))
+                for x, y in zip(donutCatalog["centroid_x"], donutCatalog["centroid_y"])
+            ]
+        )
+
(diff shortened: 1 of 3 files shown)


python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
```diff
@@ -300,6 +300,53 @@ def formatDonutCatalog(self, catalog: QTable) -> afwTable.SimpleCatalog:
...
+
+        xMin = erodedBBox.getMinX()
+        xMax = erodedBBox.getMaxX()
+        yMin = erodedBBox.getMinY()
+        yMax = erodedBBox.getMaxY()
+
+        x = np.array(donutCatalog["centroid_x"])
+        y = np.array(donutCatalog["centroid_y"])
+
+        inBounds = (x >= xMin) & (x <= xMax) & (y >= yMin) & (y <= yMax)
```
jbkalmbach: I would suggest using the `bbox.contains` function here.

tests/task/test_generateDonutFromRefitWcsTask.py
```diff
@@ -358,15 +376,87 @@ def testRefCatalogFailureWithNonExistentPhotoRefFilter(self) -> None:
...
+                    midY,  # y is fine but x is out
+                ],
+            }
+        )
+
+        clipped = self.task._clipDonutCatalogToDetector(catalog, detector)
+
+        # Only the first two sources should survive
+        self.assertEqual(len(clipped), 2)
+        self.assertAlmostEqual(clipped["centroid_x"][0], midX)
```
jbkalmbach: Could use np.testing to combine these two single tests into one array test.

tests/task/test_generateDonutFromRefitWcsTask.py
```diff
@@ -358,15 +376,87 @@ def testRefCatalogFailureWithNonExistentPhotoRefFilter(self) -> None:
...
+                "centroid_y": [
+                    erodedBBox.getMinY() + 50,
+                    erodedBBox.getMaxY() - 50,
+                    (erodedBBox.getMinY() + erodedBBox.getMaxY()) / 2,
+                ],
+            }
+        )
+
+        clipped = self.task._clipDonutCatalogToDetector(catalog, detector)
+        self.assertEqual(len(clipped), len(catalog))
```
jbkalmbach, line 464: Could test that the whole catalog before equals the whole catalog after.

tests/task/test_generateDonutFromRefitWcsTask.py
```diff
@@ -358,15 +376,87 @@ def testRefCatalogFailureWithNonExistentPhotoRefFilter(self) -> None:
...
+                ],
+            }
+        )
+
+        clipped = self.task._clipDonutCatalogToDetector(catalog, detector)
+
+        # Only the first two sources should survive
+        self.assertEqual(len(clipped), 2)
+        self.assertAlmostEqual(clipped["centroid_x"][0], midX)
+        self.assertAlmostEqual(clipped["centroid_x"][1], midY)
```
jbkalmbach: You should add a test that the log output is correct here.

jbkalmbach (approved):
Looks good. A few suggestions. Biggest one I think is testing the log message but the others you can take or leave.

suberlak:
implemented all review comments and cleaned up commit history

PR #72, opened 2026-02-25
Add ability to reselect direct detect catalog in cases of wcs fit fai…

jbkalmbach:
…lure.


suberlak (approved):
Looks good, but are there performance implications if WCS fitting fails frequently? The fallback to the original catalog might not be ideal for many consecutive failures.
Should there be a maximum retry limit or alternative catalog selection strategy if WCS fitting repeatedly fails? Have you verified this works well with different reference catalog sources and filter configurations?

jbkalmbach:
This mode of running with a fallback is already what we currently do and has run on-sky many times before. All I am adding here is the ability to trim that catalog if needed to match the characteristics we use when running source selection on the WCS catalogs.

jbkalmbach:
Tested that this successfully works on USDF data and merged

PR #77, opened 2026-03-03
Tickets/dm 53212

jbkalmbach:
This PR improves how intrinsic Zernike tables are looked up (now PrerequisiteInputs with a proper lookup function). It moves donut table pairing out of donut_viz and adds intra and extra donut table outputs to the science sensor cutout task. It adds code for a Rapid Analysis custom QG builder workflow that enables the intra and extra-focal wavefront sensors to run in parallel in RA.

This will require changes to the RapidAnalysis pipelines that are contained in the corresponding PR on `donut_viz`  (https://github.com/lsst-ts/donut_viz/pull/68) but non-RA pipelines will not need changes. This code is already the base code in `deploy-bts` and `deploy-summit` so has been tested and is already running in production.

### DIFF
--- a/python/lsst/ts/wep/task/calcZernikesTask.py
+++ b/python/lsst/ts/wep/task/calcZernikesTask.py
@@ -72,7 +72,7 @@ class CalcZernikesTaskConnections(
         storageClass="StampsBase",
         name="donutStampsIntra",
     )
-    intrinsicTables = connectionTypes.Input(
+    intrinsicTables = connectionTypes.PrerequisiteInput(
         doc="Intrinsic Zernike Map for the instrument",
         dimensions=("detector", "instrument", "physical_filter"),
         storageClass="ArrowAstropy",
--- a/python/lsst/ts/wep/task/calcZernikesTask.py
+++ b/python/lsst/ts/wep/task/calcZernikesTask.py
@@ -65,10 +65,14 @@ def lookupIntrinsicTables(
 ) -> list[DatasetRef]:
     """Assumes that the dataId is always for the extra focal at present
     """
-    ref1 = registry.findDataset(datasetType, dataId, collections=collections)
-    dataId2 = DataCoordinate.standardize(dataId, detector=dataId["detector"] + 1)
-    ref2 = registry.findDataset(datasetType, dataId2, collections=collections)
-    return [ref1, ref2]
+    detector = dataId["detector"]
+    isCornerChip = (detector in intra_focal_ids) or (detector in extra_focal_ids)
+
+    refs = [registry.findDataset(datasetType, dataId, collections=collections)]
+    if isCornerChip:  # we're running a CWFS pair, not a FAM image
+        dataId2 = DataCoordinate.standardize(dataId, detector=dataId["detector"] + 1)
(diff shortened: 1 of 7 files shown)


tests/testData/gen3TestRepo/gen3.sqlite3
suberlak, line 1: did this need to be changed?

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -72,12 +86,13 @@ class CalcZernikesTaskConnections(
         storageClass="StampsBase",
         name="donutStampsIntra",
     )
-    intrinsicTables = connectionTypes.Input(
+    intrinsicTables = connectionTypes.PrerequisiteInput(
```
suberlak, line 89: I wonder if these functionality changes are tested  in unit tests? If not, should they be ?

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
```diff
@@ -56,6 +56,19 @@ class CutOutDonutsScienceSensorTaskConnections(
         storageClass="Exposure",
         name="post_isr_image",
         multiple=True,
```
suberlak, line 58: Do these input changes mean that some backwards compatibility will break?  How was this tested?

suberlak:
Could the PR include a summary of logic changes, i.e. what are the most significant areas being modified ? Especially an overview of how science-sensors were handled before vs what does this PR introduce; I'm not sure it's clear from just code changes.

suberlak (commented):
This PR would benefit from better description of what is the issue that is being addressed / solved,  anything that would be backwards-incompatible, and possible mitigation (eg. what inputs are now needed vs what used to be sufficient)

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: They are because we run the pipelines in the unit tests.

tests/testData/gen3TestRepo/gen3.sqlite3
jbkalmbach: Yes, we add new datasets `donutTableIntra` and `donutTableExtra` and in the corresponding `donut_viz` PR we update the dataset type definition of `aggregateDonutTable`

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
jbkalmbach: Yes. The pipelines are updated in the corresponding `donut_viz` PR. This has all been running on BTS and summit already inside Rapid Analysis.

jbkalmbach: Actually for non-RA pipelines this won't change anything.

jbkalmbach: That's why we didn't even have to update the pipeline configs in the tests.

suberlak (approved):
Thanks for updated description, looks good!

PR #80, opened 2026-03-11
Tickets/DM-54371: Add blend information to stamp metadata

suberlak:
Add blend information to stamp metadata, including blend magnitude, separation, and magnitude difference, all of which are used as selection criteria.


python/lsst/ts/wep/task/donutSourceSelectorTask.py
```diff
@@ -332,8 +347,12 @@ def selectSources(
                     # when deblending. Add one to index because
                     # magDiff is all sources in magSortedDf after index=0.
                     blendMagIdx = np.where(magDiff < minMagDiff)[0] + 1
+                    blendRawIdx = blendMagIdx - 1  # indices into magDiff / blendSeparations_arr
```
jbkalmbach, line 350: I think it would help make the different arrays and indices clearer if you moved this down to above blendSeparations and blendMagDiffs and expanded the comment. As it is it took me a few read throughs of the comment and code to understand.

python/lsst/ts/wep/task/generateDonutCatalogUtils.py
```diff
@@ -204,11 +244,17 @@ def donutCatalogToAstropy(
         fieldObjects = fieldObjects[flux_sort]
         fieldObjects.meta["blend_centroid_x"] = [blendCX[idx] for idx in flux_sort]
```
jbkalmbach, line 245: With so many blend fields would it be better at this point to just move it into the donut catalog? It might make some things easier. I'm thinking for example we could avoid crawling through the metadata when aggregating for the blend information.

jbkalmbach (approved):
Looks pretty good. Left a comment about maybe now or in the future moving this blend info out of the metadata since there's a lot more now and it might help the aggregate step. Let me know your thoughts on that.

PR #88, opened 2026-04-08
Adjust bounds in danish. Use pack_params for clarity.

jbkalmbach:


python/lsst/ts/wep/estimation/danish.py
```diff
@@ -290,8 +290,24 @@ def _estimateSingleZk(
         )
 
         # Create the initial guess for the model parameters
-        x0 = [np.sum(img), 0.0, 0.0, 1.0] + [0.0] * len(nollIndices)
-        x0 += [0.0] * model.nbkg
+        x0 = model.pack_params(
+            flux=np.sum(img),
```
suberlak, line 294: Two questions - what led to the value of each these, and could some be more strict (eg. +/- inf is everything; eg. dx, dy  is `Offset in arcseconds.` so wouldn't  eg. +/- 1e3 be more than enough ?

tests/estimation/test_danish.py
```diff
@@ -74,6 +74,7 @@ def testAccuracyWithoutBinning(self) -> None:
                         "gtol": 1e-2,
                         "max_nfev": 10,
                         "verbose": 2,
+                        "x_scale": "jac",
```
suberlak, line 77: could you document in comment the addition of this kwarg? I presume it's because of the switch to `pack_params` ?

suberlak (approved):
Looks good, small change in how parameters are bound / initialized .

python/lsst/ts/wep/estimation/danish.py
jbkalmbach: These are the defaults used by Josh for danish. We could probably do some testing on other bounds but this was just a quick fix to what we were noticing in on-sky data.

tests/estimation/test_danish.py
jbkalmbach: It's part of the switch to danish v1. v1 works more consistently with this parameter set.

PR #89, opened 2026-04-08
tickets/RSO-245: Add border fraction selection to stamp selection

elanaku:


python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -207,6 +222,57 @@ def shiftCenters(self, centerArr: np.ndarray, boundary: float, distance: float)
 
         return centerArrCopy
 
+    def calculateBorderFraction(
```
jbkalmbach, line 225: Can you add a test with some toy data to `test_cutOutDonutBase` to make sure this works the way you expect?

jbkalmbach (commented):
Overall looks good, just needs a test to sanity check the `calculateBorderFraction` function. If you could add that then I'll be happy to approve.

PR #92, opened 2026-04-10
tickets/DM-54583 update the tables properly with zk deviations, intrinsics, and zernikes

PetchMa:


python/lsst/ts/wep/task/calcZernikesNeuralTask.py
```diff
@@ -1244,6 +1257,96 @@ def run(
             donutQualityTable=donutQualityTable,
         )
 
+    def _extractTartsZkVector(
+        self, donutData: dict[str, Any], key: str, fallback_key: str | None = None
+    ) -> np.ndarray:
+        """Extract self.nollIndices components from TARTS internal data."""
```
jbkalmbach: All of these new methods need full scipy format docstrings with information on the parameters and outputs.

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
```diff
@@ -1244,6 +1257,96 @@ def run(
...
+        source = donutData.get(key, None)
+        if source is None and fallback_key is not None:
+            source = donutData.get(fallback_key, None)
+        if source is None:
+            return np.full(len(self.nollIndices), np.nan, dtype=float)
+
+        fullVec = np.ravel(self._toNumpy(source)).astype(float)
+        out = np.full(len(self.nollIndices), np.nan, dtype=float)
+        for i, noll in enumerate(self.nollIndices):
+            idx = int(noll) - 4
```
jbkalmbach: Do you have a way somewhere else to make sure noll will always be >= 4? Otherwise this could cause an issue that you wouldn't know about. Maybe add a check that nollIndices is always > 4? You're essentially replicating what we have in `makeDense` but with nans.

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
```diff
@@ -1244,6 +1257,96 @@ def run(
...
+
+            rowIdx = i + 1  # index 0 is average row
+            for modeIdx, noll in enumerate(self.nollIndices):
+                zkTable[f"Z{noll}"][rowIdx] = (totalOcs[modeIdx] * u.micron).to(u.nm)
+                zkTable[f"Z{noll}_intrinsic"][rowIdx] = (intrOcs[modeIdx] * u.micron).to(u.nm)
+                zkTable[f"Z{noll}_deviation"][rowIdx] = (devOcs[modeIdx] * u.micron).to(u.nm)
+
+        # Average row from donut rows in OCS
+        avgIdx = np.where(zkTable["label"] == "average")[0][0]
+        avgTotal = np.nanmedian(np.stack(allTotalOcs, axis=0), axis=0)
```
jbkalmbach, line 1506: Why are you doing the median here instead of an average? Have you compared the results with both? Maybe you could make which one to use a configurable option to make it easier to test?

jbkalmbach (commented):
Since you are adding some new outputs I think you should add some testing that they give what you are expecting especially when dealing with coordinate systems.

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
jbkalmbach: If this is what you've been using in all your tests on sky already then it's fine. Just make a note somewhere so that we know what the `average` is in the tables (maybe an entry in the metadata of the table).

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
PetchMa: ok, I now enforce the same Noll rules as makeDense/makeSparse via checkNollIndices in validate(), and _extractTartsZkVector delegates to makeSparse with NaN padding when the TARTS vector is shorter than needed? so we do not rely on silent noll - 4 behavior for invalid modes.

python/lsst/ts/wep/task/calcZernikesNeuralTask.py
PetchMa: Yeah we want to keep nanmedian for the aggregate row (same as on-sky tests). The output zernikes table now sets meta["average_row_aggregation"] to "nanmedian" whenever that row is filled from per-donut neural components just to document things, and the method docstring notes that the ood_score aggregate in run also uses nanmedian.

jbkalmbach (approved):
Thanks for the updates. Looks good to merge.

PR #93, opened 2026-04-10
Tickets/rso 250: Update zernike clipping.

jbkalmbach:
* Add info on which zernikes were clipped in the average into zernikes metadata
* Update minimum zernike clip level to 0.1 microns.

### DIFF
--- a/python/lsst/ts/wep/task/combineZernikesSigmaClipTask.py
+++ b/python/lsst/ts/wep/task/combineZernikesSigmaClipTask.py
@@ -24,6 +24,7 @@ __all__ = ["CombineZernikesSigmaClipTaskConfig", "CombineZernikesSigmaClipTask"]
 from typing import Any
 
 import numpy as np
+import astropy.units as u
 from astropy.table import Table
 
 import lsst.pex.config as pexConfig
@@ -44,7 +45,7 @@ class CombineZernikesSigmaClipTaskConfig(CombineZernikesBaseConfig):
     )
     stdMin: pexConfig.Field = pexConfig.Field(
         dtype=float,
-        default=0.005,
+        default=0.1,
         doc="Minimum threshold for clipping the standard deviation in um.",
     )
     maxZernClip: pexConfig.Field = pexConfig.Field(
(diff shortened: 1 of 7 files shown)


python/lsst/ts/wep/task/combineZernikesSigmaClipTask.py
```diff
@@ -44,7 +45,7 @@ class CombineZernikesSigmaClipTaskConfig(CombineZernikesBaseConfig):
     )
     stdMin: pexConfig.Field = pexConfig.Field(
         dtype=float,
-        default=0.005,
+        default=0.1,
```
suberlak, line 48: I think changes to thresholds should be accompanied (eg. in the PR description or ticket comments) by some quantitative statement (eg. testing  on an average FBS night, what is the resulting increased acceptance rate? How many more donut pairs  per night do we end up using ? How much does it increase the  RMS scatter of included Zk pairs ?)

python/lsst/ts/wep/task/combineZernikesSigmaClipTask.py
```diff
@@ -94,7 +95,7 @@ def _combineZernikes(self, zkTable: Table) -> Table:
         else:
             raise ValueError(f"Unknown zkClipType: {self.zkClipType}")
         subTable = zkTable[zkTable["label"] != "average"][columns]
-        zernikeArray = np.array([subTable[col] for col in columns]).T
+        zernikeArray = np.array([subTable[col].to(u.um) for col in columns]).T
 
         # Perform conditional sigma clipping
         sigArray = conditionalSigmaClip(
```
suberlak, line 101: Could we also add `sigArray` to the metadata? That would explicitly show how much a given mode departed from the median that it got rejected.

suberlak: Although I see from this line https://github.com/lsst-ts/ts_wep/blob/daf45fc52e8c3d82648cea6dec6039e9c7aea11a/python/lsst/ts/wep/utils/miscUtils.py#L431-L433 that we only get a copy of original Zks with outliers replaced by NaNs... But I still think it would be easier to add that metadata than to try to run the `conditionalSigmaClip` on raw Zks to get the scatter that the task was dealing with.. Perhaps it would be good to store both the input and output arrays of `conditionalSigmaClip` ...

suberlak (approved):

python/lsst/ts/wep/task/combineZernikesSigmaClipTask.py
jbkalmbach: Added new metadata in the zernike tables to identify which noll indices on which donuts triggered clipping.

PR #95, opened 2026-04-16
Add temperature parameter for softmax weighting

htohfa:
Added temperature parameter for softmax weighting of predictions based on model uncertainties


htohfa:
started a new pr

PR #96, opened 2026-04-16
Add uncertainty-weighted averaging for predictions

htohfa:
Implemented uncertainty-weighted averaging for Zernike predictions, enhancing prediction accuracy based on error estimates.


python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -217,31 +231,51 @@ def _estimateZk(
...
-        # Split outputs and handle old models that don't return FWHM
+        # Split outputs and handle models with varying return values.
+        # Models may return:
+        #   - a single tensor (zk only)
+        #   - a 2-tuple (zk, zk_err)
+        #   - a 3-tuple (zk, zk_err, fwhm)
         if isinstance(outputs, tuple):
             outZk = outputs[0].cpu().numpy()
-            outFwhm = outputs[1].cpu().numpy()
+            outZk_err = outputs[1].cpu().numpy() if len(outputs) > 1 else None
```
jfcrenshaw, line 240: For consistency, you should set these to an array of NaNs, similar to the line below for outFWHM

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -217,31 +231,51 @@ def _estimateZk(
...
+            outFwhm = outputs[2].cpu().numpy() if len(outputs) > 2 else np.full((len(imgs), 2), np.nan)
         else:
             outZk = outputs.cpu().numpy()
-            outFwhm = np.full((len(imgs), 1), np.nan)
+            outZk_err = outputs[1].cpu().numpy() if len(outputs) > 1 else None
```
jfcrenshaw: This shouldn't have an if statement, and should just directly set to NaNs, like the outFWHM

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -107,7 +119,8 @@ def modelPath(self, value: str) -> None:
         """
         # Load the model
         try:
-            self.model = torch.load(value, map_location=self.device, weights_only=False)
+            self.model = torch.load(value, map_location=self.device, weights_onl
+y=False)
```
jfcrenshaw: I think there's an accidental new-line that snuck in here? `y=False` is on its own line?

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -43,6 +48,11 @@ class AiDonutAlgorithm(WfAlgorithm):
         model requirements. Default is a test model included with ts_wep.
     device : str, optional
         Device to load the model on ('cpu' or 'cuda'). Default is 'cpu'.
+    temperature : float, optional
+        Temperature parameter for softmax weighting of predictions
+        when both intra- and extra-focal images are provided. Lower
+        values place more weight on the prediction with lower
+        estimated error. Default is 1.0.
```
jfcrenshaw: The default in the doc string does not match the actual default value you set (which is `5e-3 `)

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -64,8 +74,10 @@ def __init__(
         self,
         modelPath: str = DEFAULT_MODEL_PATH,
         device: str = "cpu",
+        temperature: float = DEFAULT_TEMPERATURE,
```
jfcrenshaw, line 77: You should just directly set 0.005 as the default here. There's no reason for you to define the constant `DEFAULT_TEMPERATURE`.

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: I also expect this if statement wouldn't evaluate in the way you want. In this branch, `outputs` is almost certainly an array of Zernike coefficients, and it's `len()` is greater than 1. So it will assign to `outZk_err` the value `outputs[1].cpu().numpy()`, which will just be the value of the 2nd Zernike coefficient...

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: Also, for better or for worse (definitely for worse imo), ts_wep uses CamelCase for variable names, so you should follow this convention when adding to the code base. So you generally shouldn't use underscores in variable names. Something like `outZkErr` would be more appropriate.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -217,31 +231,51 @@ def _estimateZk(
...
         if saveHistory:
             # Metadata and I1 Zernikes
             self._history = {
                 "modelPath": self.modelPath,
                 "device": self.device,
+                "temperature": self.temperature,
                 "modelNollIndices": self.model.nollIndices,
                 I1.defocalType.value: {
                     "zk": outZk[0],
+                    "zk_err": outZk_err[0] if outZk_err is not None else None,
```
jfcrenshaw: You shouldn't have if statements here. Just `outZk_err[0]`

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -250,6 +284,7 @@ def _estimateZk(
                 self._history |= {
                     I2.defocalType.value: {
                         "zk": outZk[1],
+                        "zk_err": outZk_err[1] if outZk_err is not None else None,
```
jfcrenshaw: You shouldn't have if statements here. Just `outZk_err[1]`

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -217,31 +231,51 @@ def _estimateZk(
...
+            weights = weights / weights.sum(axis=0, keepdims=True)
+            zk = (outZk * weights).sum(axis=0)
+            
 
-        # Record average outputs
-        zk = outZk.mean(axis=0)
+        else:
+            zk = outZk.mean(axis=0)
+            print('not using weighted mean')
         zkMeta = {"fwhm": outFwhm.mean(axis=0)}
```
jfcrenshaw, line 258: You should save the weights in `zkMeta` so we can do another weighted mean via `CombineZernikes...`. The total weight here should be the sum of the weights from before, so
```python
zkMeta = {"fwhm": outFwhm.mean(axis=0), "weight": weights.sum()}
```

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -217,31 +231,51 @@ def _estimateZk(
...
+            weights = np.exp(-outZk_err / self.temperature)
+            weights = weights / weights.sum(axis=0, keepdims=True)
+            zk = (outZk * weights).sum(axis=0)
+            
 
-        # Record average outputs
-        zk = outZk.mean(axis=0)
+        else:
+            zk = outZk.mean(axis=0)
+            print('not using weighted mean')
```
jfcrenshaw: Don't put print statements in production code

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -170,7 +184,7 @@ def _estimateZk(
             Zernike coefficients for the provided Noll indices, estimated from
             the images, in meters.
         dict
-            Empty dictionary (exists for compatibility).
+            Dictionary containing metadata such as FWHM estimates.
```
jfcrenshaw, line 186: And the Zernike weight

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -64,8 +74,10 @@ def __init__(
         self,
         modelPath: str = DEFAULT_MODEL_PATH,
         device: str = "cpu",
+        temperature: float = DEFAULT_TEMPERATURE,
     ) -> None:
         self.device = device
+        self.temperature = temperature
```
jfcrenshaw, line 80: Do you see (below) how the other parameters use the property/setter paradigm to validate the value of the parameters? You should do the same for temperature. I.e., I expect it needs to be a positive float? You should check this is the case and raise a `TypeError` or `ValueError` in the appropriate circumstances.

jfcrenshaw (changes requested):
This is a good start! I left comments on your changes to `aiDonut.py`

In addition to these changes:
- You'll need to lint the files so the linting tests pass
- You need to add tests for AiDonutAlgorithm including:
	- testing bad values for temperature
	- testing the appropriate values are in zkMeta and the algorithm history
	- could you add another version of testOutputs that runs on only a single Donut? I didn't realize there wasn't an unpaired test here, and we should add that to make sure unpaired works fine. This is especially important since we're adding more model outputs and need to test that the relevant logic still works for single-donut estimation
- You need to implement a new subclass of `CombineZernikesBaseTask` that performs a weighted average for all of the Zernikes across the CCD. This can be very similar to `CombineZernikesMeanTask`, except you will need to pull the weights out of the Zernike table metadata, and replace `np.mean` with a wrapper around `np.average`  that implements your weighted average.
- You'll need to add tests for this new class

htohfa (commented):
I removed the print statements as asked. I am only doing weighted average in this code so doesnt make a lot of sense to me to return the uncertainties, I dont think the intrinsics should be weighted because they are not coming out of my model the convergence test i did was with just weighting the model zernike output

suberlak (changes requested):
Once JF's comments are addressed, I think that the  unit tests for aiDonut should be updated to test correct flow of these new arguments (eg. `temperature`) and intended behavior. Otherwise looks good!

htohfa:
> This is a good start! I left comments on your changes to `aiDonut.py`
> 
> In addition to these changes:
> 
> * You'll need to lint the files so the linting tests pass
> * You need to add tests for AiDonutAlgorithm including:
>   
>   * testing bad values for temperature
>   * testing the appropriate values are in zkMeta and the algorithm history
>   * could you add another version of testOutputs that runs on only a single Donut? I didn't realize there wasn't an unpaired test here, and we should add that to make sure unpaired works fine. This is especially important since we're adding more model outputs and need to test that the relevant logic still works for single-donut estimation
> * You need to implement a new subclass of `CombineZernikesBaseTask` that performs a weighted average for all of the Zernikes across the CCD. This can be very similar to `CombineZernikesMeanTask`, except you will need to pull the weights out of the Zernike table metadata, and replace `np.mean` with a wrapper around `np.average`  that implements your weighted average.
> * You'll need to add tests for this new class

-done, it passes all the tests
-large values would be bad temperature this is not literal temperature just weighting if you set it to 1 the relationship boils down to e^-uncertainty this is mathematical there's no ambiguity 
-unpaired test works 
- we process one corner at a time then it sees the stamps and does the weighted averaging i am not sure why we need this new class

jfcrenshaw:
You've marked a lot of the comments as resolved, but the requested changes don't appear in the PR. Perhaps you have commits you've forgotten to push?

> I am only doing weighted average in this code so doesnt make a lot of sense to me to return the uncertainties, I dont think the intrinsics should be weighted because they are not coming out of my model the convergence test i did was with just weighting the model zernike output

You have currently only implemented weighted averaging within `AiDonutAlgorithm`, and you can see from [the docstring](https://github.com/lsst-ts/ts_wep/blob/daf45fc52e8c3d82648cea6dec6039e9c7aea11a/python/lsst/ts/wep/estimation/aiDonut.py#L150-L153) this only estimates Zernikes for one or two donuts at a time. `CalcZernikesTask` will run this algorithm on all of the donuts (or donut pairs) for the CCDs, and will then use the `CombineZernike...` subtasks to average Zernikes across the CCD. With your current limited implementation, there are no weights used when averaging across the CCDs.

When averaging across the CCDs, it is important that we average the OPD, the intrinsics, and the wavefront deviation using the exact same scheme. The averaged intrinsics tell us what the effective intrinsics were for that CCD in the exposure. This is an important data product. This kind of thing is totally independent of which wavefront estimator we use, whether it's AiDonut, Danish, or TIE. All of the intrinsics handling happens downstream from AiDonut.


> large values would be bad temperature this is not literal temperature just weighting if you set it to 1 the relationship boils down to e^-uncertainty this is mathematical there's no ambiguity

You misunderstand - I know this is not a physical temperature. However, setting `temperature="this is a string, not a number!"` would be bad, because you should not set the temperature to a string (or a list, or a dictionary, or a None, or ...). Furthermore, setting `temperature=-1` is probably undesirable because it will *upweight* rather than *downweight* high uncertainties. Most (but not all!) of these cases should cause transparent errors at some stage in the pipeline, but it is better practice to have explicit type/value checks.

This probably seems overly pedantic, but it's important to remember that `ts_wep` is not *research* code - it is *production* code that runs inside an important pipeline that connects to a large and complicated software ecosystem. It's important to have explicit checks and clear error messages because things can go wrong in quite unexpected ways, and it's very helpful to know, for example, "oh for some reason this other part of the pipeline is trying to pass a string where it should be passing a float". That's not an imagined scenario - it does happen!

> it passes all the tests

You're missing the new tests though. Perhaps these are in commits you forgot to push?

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: Marking this as unresolved. Not sure if you fixed this in a commit that didn't make it to github?

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: Marking this as unresolved. Not sure if you fixed this in a commit that didn't make it to github?

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -105,27 +121,26 @@ def modelPath(self, value: str) -> None:
...
-        - "nollIndices" - Noll indices for which Zernikes are returned
-        - "zk" - the final, averaged zernike estimate
+
+        - ``"modelPath"`` - path to the PyTorch model file
+        - ``"device"`` - device used for inference
+        - ``"temperature"`` - temperature used for weighted averaging
+        - ``"modelNollIndices"`` - Noll indices the model predicts
+        - ``"intra"`` and/or ``"extra"`` - all Zernikes returned by the model
+        - ``"nollIndices"`` - Noll indices for which Zernikes are returned
+        - ``"zk"`` - the final, averaged Zernike estimate
```
jfcrenshaw: Just wondering, why did you add the ``'s around the quotes here?

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -142,122 +157,141 @@ def _estimateZk(
...
+
+        # Compute weighted average using model uncertainties if available,
+        # otherwise fall back to a simple mean.
+        # Weighting uses softmax(-err / temperature) so that predictions
+        # with lower estimated error receive higher weight.
+        if out_zk_err is not None:
+            weights = np.exp(-out_zk_err / self.temperature)
+            weights = weights / weights.sum(axis=0, keepdims=True)
+            zk = (out_zk * weights).sum(axis=0)
+            zk_err_avg = (out_zk_err * weights).sum(axis=0)
```
jfcrenshaw: I'm confused by this definition. Why do we want to define a weighted error like this? Is it useful? Plus this is not the usual formula for the standard error of a weighted mean.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -142,122 +157,141 @@ def _estimateZk(
...
-        zk = outZk.mean(axis=0)
-        zkMeta = {"fwhm": outFwhm.mean(axis=0)}
+        zk_meta = {"fwhm": out_fwhm[:, 0].mean(), "zk_err": zk_err_avg}
 
-        # Only return requested nollIndices
+        # Sparsify zk and zk_err to the requested Noll indices.
         zk = makeDense(zk, self.model.nollIndices)
         zk = makeSparse(zk, nollIndices)
+        zk_err_dense = makeDense(zk_err_avg, self.model.nollIndices)
+        zk_err_sparse = makeSparse(zk_err_dense, nollIndices)
```
jfcrenshaw: Does `zk_err_avg` contain a value per Zernike coefficient? I had thought you said the errors were a single scalar per donut, rather than a scalar per Zernike per donut?

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -142,122 +157,141 @@ def _estimateZk(
...
-            # Average Zernikes
             self._history |= {
                 "nollIndices": nollIndices,
                 "zk": zk,
-                "fwhm": zkMeta["fwhm"],
+                "fwhm": zk_meta["fwhm"],
             }
 
-        return zk, zkMeta
+        return zk, zk_meta
```
jfcrenshaw: Why are you renaming variables from CamelCase to snake_case?

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: Marking this as unresolved. Not sure if you fixed this in a commit that didn't make it to github?

python/lsst/ts/wep/task/combineZernikesWeightedTask.py
jfcrenshaw, line 1: Why are you using the Zernike errors to do weighting like this, compared to in `aiDonut.py` where you use a temperature to convert the errors into weights? I think it will be very confusing to use the same values in two totally different weighting schemes.

I think it probably makes more sense to store the weights in zkMeta, rather than the model's "zernike errors". Then this class can just load those weights and do a weighted mean. This will also have the benefit of being more abstracted, in that it can work with any model that returns weights.

python/lsst/ts/wep/task/combineZernikesWeightedTask.py
```diff
@@ -0,0 +1,63 @@
...
+    ----------
+    *args : object
+        Positional arguments passed to the parent class.
+    **kwargs : object
+        Keyword arguments passed to the parent class.
+    """
+
+    def _combineZernikes(self, zkTable: Table) -> Table:
+        """Combine Zernike coefficients using a weighted average.
+        Parameters
```
jfcrenshaw, line 73: Add a blank line above "Parameters"

python/lsst/ts/wep/task/combineZernikesWeightedTask.py
```diff
@@ -0,0 +1,63 @@
...
+        ----------
+        zkTable : `astropy.table.Table`
+            Table of Zernike coefficients to combine.
+        Returns
```
jfcrenshaw, line 78: Add a blank line above "Returns"

python/lsst/ts/wep/task/combineZernikesWeightedTask.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class CombineZernikesWeightedTask(CombineZernikesSigmaClipTask):
+    """Like CombineZernikesSigmaClipTask but uses weighted average.
+    Parameters
+    ----------
+    *args : object
+        Positional arguments passed to the parent class.
+    **kwargs : object
+        Keyword arguments passed to the parent class.
```
jfcrenshaw: You can remove these lines from the docstring

python/lsst/ts/wep/task/combineZernikesWeightedTask.py
```diff
@@ -0,0 +1,63 @@
+__all__ = ["CombineZernikesWeightedTask"]
+import numpy as np
+from astropy.table import Table
+
+from lsst.ts.wep.task.combineZernikesSigmaClipTask import CombineZernikesSigmaClipTask
+
+
```
jfcrenshaw, line 67: You will probably need to create a new config class for this task even though it doesn't define any new configs beyond what's available in the `CombineZernikesSigmaClipTask` parent task.

It should probably look like
```python
class CombineZernikesWeightedTaskConfig(CombineZernikesSigmaClipTaskConfig):
     """Config for combining Zernikes via a weighted average with  sigma clipping."""
     ...
```

I've asked on slack in #dm-newbies if this is the correct pattern for this, so I'll follow up when I get an answer

jfcrenshaw:
Good progress!

I added new comments, and re-opened a few of the previous "resolved" comments that didn't seem to be resolved.

We're still missing new tests.

We need these tests for `AiDonutAlgorithm`:
> You need to add tests for AiDonutAlgorithm including:
> - testing bad values for temperature
> - testing the appropriate values are in zkMeta and the algorithm history
> - could you add another version of testOutputs that runs on only a single Donut? I didn't realize there wasn't an unpaired test here, and we should add that to make sure unpaired works fine. This is especially important since we're adding more model outputs and need to test that the relevant logic still works for single-donut estimation

And you'll need to write a new test file for the new `CombineZernikesWeightedTask`. You can just copy the test file for one of the other `CombineZernikes...` subclasses, and adapt these tests to match your new class.

python/lsst/ts/wep/estimation/aiDonut.py
htohfa: line 1: this weights the more uncertain samples even lower with the temperature 

line 2: we want the weight to be normalized to 1

line 3: doing weighted average with the normalized weight

line 4: averaging over the zk error in case you're running only 2 donuts at a time and want to use this error for some later averaging you asked that we propagate the error for averaging over a detector. If you have say 4 pairs of donuts for one detector and you want to average over that, without this you will have one zernike and two zk_error per pair. This gives you an averaged error for this zernike error that you can use

python/lsst/ts/wep/estimation/aiDonut.py
htohfa: i do output zernike weight now not sure what's unresolved here?

python/lsst/ts/wep/task/combineZernikesWeightedTask.py
jfcrenshaw: Okay, I checked with the DM gurus, and they said we don't need to redefine a new config class like this, so you can ignore this comment. I will mark as resolved.

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: I was asking about your "line 4" (labeled line 256 in the snippet). My issue still isn't resolved, but since this reference is now outdated I'll close this and re-open a new comment on the up-to-date file.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -193,71 +222,95 @@ def _estimateZk(
...
+        else:
+            pairWeight = 1.0
+            weights = np.full_like(outZk, 1.0 / len(imgs))
+            zk = outZk.mean(axis=0)
+            zkErrAvg = np.full(zk.shape, np.nan)
 
-        # Record average outputs
-        zk = outZk.mean(axis=0)
-        zkMeta = {"fwhm": outFwhm.mean(axis=0)}
+        zkMeta = {"fwhm": outFwhm.mean(axis=0), "zk_err": zkErrAvg, "weight": pairWeight}
```
jfcrenshaw: Sometimes you use the key `"zk_err"`, and other times you use the key `"zkErr"`. Can you standardize this? I'd recommend changing these all to `"zkErr"`.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -193,71 +222,95 @@ def _estimateZk(
...
 
-        # Only return requested nollIndices
+        # Sparsify zk and zkErr to the requested Noll indices.
         zk = makeDense(zk, self.model.nollIndices)
         zk = makeSparse(zk, nollIndices)
+        zkErrDense = makeDense(zkErrAvg, self.model.nollIndices)
+        zkErrSparse = makeSparse(zkErrDense, nollIndices)
+        zkMeta["zk_err"] = zkErrSparse
```
jfcrenshaw: Sometimes you use the key `"zk_err"`, and other times you use the key `"zkErr"`. Can you standardize this? I'd recommend changing these all to `"zkErr"`.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -193,71 +222,95 @@ def _estimateZk(
...
+                        "zkErr": outZkErr[1],
                         "fwhm": outFwhm[1],
                     }
                 }
-            # Average Zernikes
             self._history |= {
                 "nollIndices": nollIndices,
                 "zk": zk,
                 "fwhm": zkMeta["fwhm"],
+                "zk_err": zkMeta["zk_err"],
```
jfcrenshaw: Sometimes you use the key `"zk_err"`, and other times you use the key `"zkErr"`. Can you standardize this? I'd recommend changing these all to `"zkErr"`.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -193,71 +222,95 @@ def _estimateZk(
...
+            outZkErr = np.full_like(outZk, np.nan)
+            outFwhm = np.full((len(imgs), 2), np.nan)
+
+        # Compute weighted average using model uncertainties if available,
+        # otherwise fall back to a simple mean.
+        # Weighting uses softmax(-err / temperature) so that predictions
+        # with lower estimated error receive higher weight.
+        if np.all(np.isfinite(outZkErr)):
+            rawWeights = np.exp(-outZkErr / self.temperature)
+            pairWeight = float(rawWeights.mean())
```
jfcrenshaw: This won't matter computationally since you end up renormalizing in the `CombineZernikesWeightedTask`, but the pairWeights really should be the sum of the weights (i.e. `rawWeights.sum()`) instead of the mean.

This can be seen if you consider four measurements $x_i$ with associated weights $w_i$. If we just compute the weighted mean of all four measurements, we of course have

$$
\bar{x} = \frac{w_1 x_1 + w_2 x_2 + w_3 x_3 + w_4 x_4}{w_1 + w_2 + w_3 + w_4}
$$

Instead imagine, however, that we arrange these points into two sets of pairs, $A = (x_1, x_2)$ and $B = (x_3, x_4)$, and we first calculate the pair averages:

$$
\bar{x}_A = \frac{w_1 x_1 + w_2 x_2}{w_1 + w_2} \quad \text{and} \quad \bar{x}_B = \frac{w_3 x_3 + w_4 x_4}{w_3 + w_4}.
$$

Now we want to combine these two averages into our final average such that

$$
\bar{x} = \frac{w_A \bar{x}_A + w_B \bar{x}_B}{w_A + w_B}.
$$

It's easy to show the correct choice is the sum of the component weights

$$
w_A = w_1 + w_2 \quad \text{and} \quad w_B = w_3 + w_4.
$$

In particular,

$$
\bar{x}
= \frac{w_A \bar{x}_A + w_B \bar{x}_B}{w_A + w_B}
= \frac{(w_1 + w_2) \frac{w_1 x_1 + w_2 x_2}{w_1 + w_2}  + (w_3 + w_4) \frac{w_3 x_3 + w_4 x_4}{w_3 + w_4}}{w_1 + w_2 + w_3 + w_4}
= \frac{w_1 x_1 + w_2 x_2 + w_3 x_3 + w_4 x_4}{w_1 + w_2 + w_3 + w_4}
$$

as desired.

Note the conclusion is invariant under the redefinition $(w_A, w_B) \to (\frac{w_A}{c}, \frac{w_B}{c}$ for some scalar $a$. This is an accident of the fact that we started with two pairs, however. If set $B$ consisted of a single donut (or a triplet), this symmetry would be broken, indicating that the sum is the unique correct choice.

So, again, you should do `rawWeights.sum()` instead of `rawWeights.mean()`. This difference shouldn't really matter in the end because our sub-averages should always consist of either _only_ individual donuts, or _only_ donut pairs, so the symmetry described above shouldn't be broken. Regardless, we should make sure to do the correct thing here and take the sum rather than the mean.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -193,71 +222,95 @@ def _estimateZk(
...
+            weights = rawWeights / rawWeights.sum(axis=0, keepdims=True)
+            zk = (outZk * weights).sum(axis=0)
+            zkErrAvg = (outZkErr * weights).sum(axis=0)
```
jfcrenshaw: I think I have some qualms about defining an average "error" like this.

If these really are errors, this is not how you compute a weighted error because the weighted mean is a linear combination of the individual predictions and the errors should therefore add in quadrature. I.e., you should really do

```python
zkErrAvg = np.sqrt((outZkErr**2 * weights**2).sum(axis=0))
```

 This assumes, however, that `zkErrAvg` actually can be interpreted as the error of the Zernike estimates, which presumably it cannot since you are using softmax weights instead of inverse variance weights. Under this schema, it makes more sense to tie the "average error" directly to the average weight, i.e.

$$
e_\mathrm{avg} \equiv -T ~ \ln[\texttt{pairWeight}]
$$

so that the pair weight is, consistently,

$$
w_\text{avg} = \exp \left( -\frac{e_\mathrm{avg}}{T} \right)
=  \exp \left( -\frac{-T ~ \ln[\texttt{pairWeight}]}{T} \right) 
= \texttt{pairWeight},
$$

as desired.

However, we don't ever use this "average error", so we don't really need to compute and store it, so a second option is to just delete these lines.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -193,71 +222,95 @@ def _estimateZk(
...
 
-        # Split outputs and handle old models that don't return FWHM
+        # Split outputs. Models may return:
+        #   - a single tensor (zk only)
+        #   - a 2-tuple (zk, zkErr)
+        #   - a 3-tuple (zk, zkErr, fwhm)
         if isinstance(outputs, tuple):
             outZk = outputs[0].cpu().numpy()
-            outFwhm = outputs[1].cpu().numpy()
+            outZkErr = outputs[1].cpu().numpy() if len(outputs) > 1 else np.full_like(outZk, np.nan)
```
jfcrenshaw: Sorry to do this, but I would feel better if we called this `zkScore` instead of `zkErr` since these values cannot be consistently interpreted as the error of the measurement. Naming this `zkErr` could be confusing in the long run (see related comments about the weighted averaging).

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
```diff
@@ -44,6 +42,13 @@ class EstimateZernikesAiDonutConfig(EstimateZernikesBaseConfig):
         default="cpu",
         doc="Device to run the model on ('cpu' or 'cuda').",
     )
+    temperature: pexConfig.Field = pexConfig.Field(
+        dtype=float,
+        default=0.005,
+        doc="Temperature for softmax weighting of predictions based on model "
+        "uncertainty. Lower values trust lower-uncertainty predictions "
+        "more aggressively.",
```
jfcrenshaw: The statement "Lower values trust lower-uncertainty predictions more aggressively" is confusing to me. I'd suggest replacing it with "lower values put greater weight on lower-uncertainty predictions".

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
```diff
@@ -58,10 +63,8 @@ def wfAlgoName(self) -> WfAlgorithmName:
 
     def _applyToList(self, fun: Callable, args: Iterable, numCores: int) -> list:
         """Apply a function to a list of arguments, optionally in parallel.
-
         Overriding this method from the base class to avoid multiprocessing
         issues with AiDonut.
-
         Parameters
```
jfcrenshaw, line 72: Please put these new lines back into the docstring. You should not blindly use AI to reformat files!

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
```diff
@@ -18,9 +18,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
-
 __all__ = ["EstimateZernikesAiDonutConfig", "EstimateZernikesAiDonutTask"]
-
 from typing import Callable, Iterable
```
jfcrenshaw, line 24: Please put these new lines back. You should not blindly use AI to reformat files!

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
```diff
@@ -71,13 +74,11 @@ def _applyToList(self, fun: Callable, args: Iterable, numCores: int) -> list:
         numCores : int
             The number of cores to use. Must be 1 for AiDonut. If not 1,
             we will still run, but will raise a warning.
-
         Returns
```
jfcrenshaw, line 82: Sorry to keep repeating it, but please put the blank line back!

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
```diff
@@ -71,13 +74,11 @@ def _applyToList(self, fun: Callable, args: Iterable, numCores: int) -> list:
...
         -------
         list
             A list of results from applying the function to the arguments.
         """
         if numCores != 1:
             self.log.warn("AiDonut does not support multiprocessing. Running on a single core.")
-
         return [fun(arg) for arg in args]
```
jfcrenshaw, line 90: Sorry to keep repeating it, but please put the blank line back!

tests/estimation/test_aidonut.py
```diff
@@ -18,13 +18,11 @@
...
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
-
 import unittest
 
 from lsst.ts.wep.utils.testUtils import enforce_single_threading
 
 enforce_single_threading()
-
 from lsst.ts.wep.estimation import AiDonutAlgorithm  # noqa: E402
```
jfcrenshaw, line 28: Again, please put these blank lines back

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -105,27 +136,26 @@ def modelPath(self, value: str) -> None:
...
             Path to the PyTorch model file.
         """
-        # Load the model
         try:
             self.model = torch.load(value, map_location=self.device, weights_only=False)
         except FileNotFoundError as e:
             raise FileNotFoundError(f"Model file not found: {value}") from e
-        self.model.eval()  # Put in evaluation mode
-
-        # If loading the model succeeded, save the path
```
jfcrenshaw, line 115: Please don't remove comments like these - they're helpful for understanding the code.

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: For future reference: the model returns a single value, but rather than returning it as a float, it returns an array full of that value, where the array shape matches the Zernike shape.

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: Done.

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
jfcrenshaw: Done.

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
jfcrenshaw: Done.

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
jfcrenshaw: Done.

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
jfcrenshaw: Done.

python/lsst/ts/wep/task/estimateZernikesAiDonutTask.py
jfcrenshaw: Done.

tests/estimation/test_aidonut.py
jfcrenshaw: Done.

python/lsst/ts/wep/estimation/aiDonut.py
jfcrenshaw: Done.

jfcrenshaw (approved):
LGTM

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -91,6 +102,26 @@ def device(self, value: str) -> None:
...
+    @temperature.setter
+    def temperature(self, value: float) -> None:
+        """Set the temperature parameter.
+
+        Parameters
+        ----------
+        value : float
+            Temperature parameter. Must be a positive float.
+        """
+        if not isinstance(value, (int, float)):
```
suberlak: The check accepts int or float: isinstance(value, (int, float)).
But the error message says "Temperature must be a float." — which contradicts the check, since ints are allowed.

suberlak: eg. 
```if not isinstance(value, (int, float)) or isinstance(value, bool):
    raise TypeError("Temperature must be a number.")
if value <= 0:
    raise ValueError("Temperature must be positive.")
self._temperature = float(value)```

File: python/lsst/ts/wep/estimation/aiDonut.py [hunk 34]
```diff
@@ -193,71 +222,88 @@ def _estimateZk(
...
+                focalFlagsTch,
+                bandsTch,
             )
 
-        # Split outputs and handle old models that don't return FWHM
+        # Split outputs. Models may return:
+        #   - a single tensor (zk only)
+        #   - a 2-tuple (zk, zkScore)
+        #   - a 3-tuple (zk, zkScore, fwhm)
         if isinstance(outputs, tuple):
```
suberlak, line 271: Any existing model that returned (zk, fwhm) as a 2-tuple will now have its FWHM silently reinterpreted as zkScore, which then drives the weighting. The model requirements docstring doesn't mention zkScore , so the contract change isn't documented. if this is a desired change, it's worth to 
require named outputs (e.g. a dict or NamedTuple) so position doesn't matter. At minimum, document this as a breaking change (since now  any 2-tuple is not treated as (zk, fwhm) but  (zk, score) ...

tests/task/test_combineZernikesWeightedTask.py
```diff
@@ -0,0 +1,96 @@
...
+    def testCombineZernikes(self) -> None:
+        nPairs = 10
+        inTable = _makeTable(nPairs, np.ones(nPairs))
+        outTable = self.task.combineZernikes(inTable)
+        self.assertTrue(all(outTable["used"]))
+        for j, col in enumerate(ALL_COLS):
+            expected = float(j)
+            self.assertAlmostEqual(float(outTable[outTable["label"] == "average"][col][0]), expected)
+
+    def testCombineZernikesVaryingWeights(self) -> None:
```
suberlak, line 68: If I understand it correctly, `testCombineZernikesVaryingWeights` looks like it should verify that weights affect the average — it sets weights[9] = 1.0, all others zero. But in `_makeTable`, every pair gets the same value for a given column (`vals = list(np.zeros(nPairs) + j`)). So the weighted average is j regardless of the weights, and so is the unweighted average. The assertion would pass even if the weighting code  just returned `np.nanmean`. Eg for Z4, it asserts the average is 0.0. For Z5, 1.0. Etc.   But  since pair 9's value for Z4 is 0.0, as well as pair 0 or pair 5. So is the unweighted mean. So is the weighted mean with any weights at all. Every pair has the same value, so any way of combining them gives back j. The assertion has no power to distinguish weighted from unweighted. To actually test weighting, the pairs need to have different values, so that the weighted mean and the unweighted mean give different answers. Something like: 
``` for j, col in enumerate(ALL_COLS):
    # pair i in column j has value j + i, so pairs differ from each other
    vals = [float(j + i) for i in range(nPairs)]
    cols[col] = [-1.0] + vals
```
Now for Z4 (j=0): pair0=0, pair1=1, pair2=2, ..., pair9=9. That way, unweighted mean: (0+1+...+9)/10 = 4.5, and 
weighted mean with all weight on pair 9: 9.0 . Those are different, so the assertion assertAlmostEqual(avg, 9.0) actually tests that weighting happened. If someone deletes the weighting logic, the test fails with 4.5 != 9.0.

python/lsst/ts/wep/estimation/aiDonut.py
```diff
@@ -193,71 +222,88 @@ def _estimateZk(
...
             outZk = outputs.cpu().numpy()
-            outFwhm = np.full((len(imgs), 1), np.nan)
+            outZkScore = np.full_like(outZk, np.nan)
+            outFwhm = np.full((len(imgs), 2), np.nan)
+
+        # Compute weighted average using model uncertainties if available,
+        # otherwise fall back to a simple mean.
+        # Weighting uses softmax(-score / temperature) so that predictions
+        # with lower estimated error receive higher weight.
+        if np.all(np.isfinite(outZkScore)):
```
suberlak: `np.all(np.isfinite(outZkScore))` returns True only if every single element of the score array is finite. If even one is NaN or inf, the whole thing falls back to a plain mean.
If the model returns predictions for an intra/extra pair, is it ever possible for eg. extra-focal score to  come back as NaN (numerical issue, edge case, etc) ? Then the intra-focal score is a perfectly good number — eg. 0.01, indicating high confidence. Under the current logic, that information is thrown away entirely and both predictions are averaged equally, even though we have a strong reason to weight the intra prediction more. Did you consider a more granular alternative, eg weight the predictions whose scores are finite, and either drop or assign a default weight to the ones that aren't ? Worth adding a comment

python/lsst/ts/wep/task/combineZernikesWeightedTask.py
```diff
@@ -0,0 +1,112 @@
...
+        super()._combineZernikes(zkTable)
+
+        # Redo the average using weights over the sigma-clipped survivors.
+        use_idx = np.where(zkTable[zkTable["label"] != "average"]["used"])[0]
+
+        estimator_info = zkTable.meta.get("estimatorInfo", {})
+        weight = estimator_info.get("weight", None)
+        if weight is not None:
+            weight = np.array(weight)
+            if weight.ndim == 2:
```
suberlak: What estimator produces a 2D weight, and why is mean(axis=1) the right way to collapse it?

python/lsst/ts/wep/estimation/aiDonut.py
suberlak: Also here: `isinstance(True, int) == True`  since bool is a subclass of int. So eg. `AiDonutAlgorithm(temperature=True)` would store `1.0`, and `AiDonutAlgorithm(temperature=False)` would raise a ValueError. To be defensible against that could do 
```
if isinstance(value, bool) or not isinstance(value, (int, float)):
    raise TypeError("Temperature must be a number.")
```   that way both "True" and "False" will raise ValueError

**Review by suberlak — COMMENTED** [review]
Few small comments

File: python/lsst/ts/wep/estimation/aiDonut.py (hunk 36 above)
Comment (htohfa) [reply python/lsst/ts/wep/estimation/aiDonut.py]: changed to this
        # Zero out entire stamp if any of its scores are NaN — a NaN score
        # for one Zernike indicates the whole stamp prediction is unreliable.
        # This allows the other stamp in the pair to still contribute normally.
        finite_mask = np.isfinite(outZkScore).all(axis=1)  # (N,) per stamp
        if finite_mask.any():
            rawWeights = np.where(
                finite_mask[:, None],
                np.exp(-outZkScore / self.temperature),
                0.0,
            )
            pairWeight = float(rawWeights.sum())
            col_sums = rawWeights.sum(axis=0, keepdims=True)
            weights = np.where(col_sums > 0, rawWeights / col_sums, 1.0 / finite_mask.sum())
            zk = (outZk * weights).sum(axis=0)
        else:
            pairWeight = 1.0
            zk = outZk.mean(axis=0)

File: python/lsst/ts/wep/task/combineZernikesWeightedTask.py (hunk 37 above)
Comment (htohfa) [reply python/lsst/ts/wep/task/combineZernikesWeightedTask.py]: got rid of it, this was from a version i was writing when i allowed intra and extra pair to output seperate weight

File: python/lsst/ts/wep/estimation/aiDonut.py (hunk 34 above)
Comment (htohfa) [reply python/lsst/ts/wep/estimation/aiDonut.py:271]: fixed the comment

File: python/lsst/ts/wep/estimation/aiDonut.py (hunk 33 above)
Comment (htohfa) [reply python/lsst/ts/wep/estimation/aiDonut.py]: changed to    
     if isinstance(value, bool) or not isinstance(value, (int, float)):
            raise TypeError("Temperature must be a number.")

File: python/lsst/ts/wep/task/combineZernikesWeightedTask.py (hunk 37 above)
Comment (suberlak) [reply python/lsst/ts/wep/task/combineZernikesWeightedTask.py]: Perhaps given https://rubin-obs.slack.com/archives/C07QM715AJY/p1777572867544479  we should actually allow that ... @jfcrenshaw what do you think ?

**Review by suberlak — COMMENTED** [review]
Just noticed: 
`combineZernikesWeightedTask.py` still uses snake_case (use_idx, estimator_info, zk_columns, _weightedMean) while the edited aiDonut.py was converted to camelCase - was that intentional?

File: tests/task/test_combineZernikesWeightedTask.py [hunk 38]
```diff
@@ -0,0 +1,95 @@
...
+
+    def testCombineZernikesWithRejection(self) -> None:
+        nPairs = 10
+        inTable = _makeTable(nPairs, np.ones(nPairs), outlierPairIdx=0)
+        outTable = self.task.combineZernikes(inTable)
+        pairUsed = outTable[outTable["label"] != "average"]["used"]
+        self.assertFalse(pairUsed[0])
+        self.assertTrue(all(pairUsed[1:]))
+
+    def testTaskRun(self) -> None:
```
suberlak, line 87: `testTaskRun` got weaker. The previous version had a per-column assertion checking the average values; the new version dropped those assertions; an easy fix: would be to add back the per-column check using `expected = float(j) + 4.5` like in `testCombineZernikes`...

python/lsst/ts/wep/task/__init__.py
```diff
@@ -23,4 +24,4 @@
 from .pairTask import *
 from .reassignCwfsCutoutsTask import *
 from .refCatalogInterface import *
-from .calcZernikesNeuralTask import *
+from .calcZernikesNeuralTask import *
```
suberlak: Trailing newline  is missing here  - was it intentional?

python/lsst/ts/wep/estimation/aiDonut.py
suberlak: Ok, the 2-tuple semantics change still undocumented as a breaking change. The model-requirements docstring lists what the model must accept and output, but it doesn't say "if you return a 2-tuple, the second element is now interpreted as zkScore, not fwhm." Any existing model returning (zk, fwhm) will silently have its FWHM treated as a confidence score. The towncrier fragment also doesn't flag this. Worth eg. 
* updating the docstring to enumerate the supported output formats explicitly, or
* bump to a 3-tuple-required contract and make 2-tuples mean (zk, zkScore) only by convention going forward.

suberlak:
Once these minor comments get addressed it looks good!

python/lsst/ts/wep/task/__init__.py
htohfa: addded the line back

tests/task/test_combineZernikesWeightedTask.py
htohfa: added this:    
 for j, col in enumerate(ALL_COLS):
        expected = float(j) + 4.5
        self.assertAlmostEqual(float(outTable[outTable["label"] == "average"][col][0]), expected)

python/lsst/ts/wep/estimation/aiDonut.py
htohfa: added the it to the docstring 
- Must output one of:

        - a tensor of shape (N, n_zernikes), where n_zernikes is the
          number of Zernike coefficients predicted, starting with Noll
          index 4.
        - a 2-tuple ``(zk, zkScore)``, where ``zk`` is the tensor above
          and ``zkScore`` is a tensor of shape (N, n_zernikes) giving a
          per-coefficient confidence score (lower is better). ``zkScore``
          is used for softmax weighting between intra/extra predictions.
        - a 3-tuple ``(zk, zkScore, fwhm)``, where ``fwhm`` is a tensor
          of shape (N, 2) giving FWHM estimates for each stamp.

    - Zernikes must be returned in meters.
    - Must have a ``nollIndices`` attribute listing the Noll indices
      corresponding to the output coefficients.

    .. warning::

        This is a breaking change from earlier versions of this algorithm,
        where a 2-tuple output was interpreted as ``(zk, fwhm)``. The
        second element of a 2-tuple is now ``zkScore``, not ``fwhm``.
        Models that previously returned ``(zk, fwhm)`` must be updated
        to return either ``zk`` alone or the 3-tuple ``(zk, zkScore, fwhm)``.

also updated the fragment to reflect this

suberlak (approved):
Thanks for updating the towncrier and module code - in combination, the release notes for people upgrading, API docs for people reading the code, and a prominent warning for anyone who might miss both, ensures that it won't be left unnoticed that the 2-tuple contract for content of `output` has been changed. Also, implemented a bug fix in the partial-NaN handling, a test that now actually tests what it claims to, type-validation edge cases handled,   style consistency restored.  Looks good to merge! Let's test it on BTS asap.

PR #100, opened 2026-04-24
DM-54784: prevent least_squares ValueError

suberlak:


python/lsst/ts/wep/estimation/danish.py
```diff
@@ -305,6 +305,44 @@ def _estimateSingleZk(
             instrument=instrument,
         )
 
+        # Early bail-out for bad images.
```
jbkalmbach: It would be a lot easier to maintain if you combine this with the `try...except` down below. Maybe you could add a `raise ValueError` inside the try if the sum of the `img` < 0. Then we won't have all this repeated code.

jbkalmbach (approved):
Thanks for reconfiguring. Looks nice and clean.

PR #103, opened 2026-05-07
Tickets/DM-54890: fix DonutCorrelator

suberlak:
Guard correlator against  a situation when the entire image is masked.


jbkalmbach (commented):
Needs a quick test added to `test_donutSizeCorrelator.py`

jbkalmbach (approved):
LGTM

PR #106, opened 2026-05-10
DM-54906: Add support for danish v1.1 features

jmeyers314:
New danish features include:
  - Angle-of-incidence throughput modeling
  - Improved loss function to handle model discrepancy residuals.

### DIFF
--- a/python/lsst/ts/wep/estimation/observingConditions.py
+++ b/python/lsst/ts/wep/estimation/observingConditions.py
@@ -34,7 +34,9 @@ class ObservingConditions:
     Parameters
     ----------
     rtp : Angle or None, optional
-        Rotation angle of the camera on the telescope.
+        Rotation angle of the camera on the telescope.  Note that this is
+        not the same as BORESIGHT_ROT_ANG, which is the rotation angle of
+        the camera on the sky.
         (the default is None, meaning unknown or not applicable)
     altitude : Angle or None, optional
         Boresight altitude.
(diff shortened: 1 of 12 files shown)


python/lsst/ts/wep/estimation/observingConditions.py
```diff
@@ -0,0 +1,45 @@
...
+
+@dataclass
+class ObservingConditions:
+    """Per-observation telescope pointing state for wavefront estimation.
+
+    Parameters
+    ----------
+    rtp : Angle or None, optional
+        Rotation angle of the camera on the telescope.
+        (the default is None, meaning unknown or not applicable)
```
jbkalmbach, line 40: I think it might be helpful to include that this is not equivalent to BORESIGHT_ROT_ANG

python/lsst/ts/wep/task/estimateZernikesBase.py
```diff
@@ -168,28 +172,36 @@ def _applyToList(fun: Callable, args: Iterable, numCores: int) -> list:
...
+        rtp = None
+        altitude = None
         try:
             rsp = metadata["BORESIGHT_ROT_ANGLE_RAD"]
             q = metadata["BORESIGHT_PAR_ANGLE_RAD"]
+            rtp = Angle(q - rsp - np.pi / 2, "rad")
         except KeyError:
-            return Angle(np.nan, "rad")
-        return Angle(q - rsp - np.pi / 2, "rad")
+            pass
```
jbkalmbach: Can you include a log message for each of the errors?

jbkalmbach (approved):
Looks good. A couple small comments.

PR #112, opened 2026-05-26
Switch calcZernikeTask to using ip_isr calibration for intrinsicZernikes

gmegh:


python/lsst/ts/wep/bin/ingestIntrinsicZernikes.py
```diff
@@ -0,0 +1,190 @@
...
+        help="Dataset type to write to. (default: intrinsicZernikes)",
+        metavar="TYPE",
+    )
+    parser.add_argument(
+        "--output-run",
+        type=str,
+        required=True,
+        help=(
+            "RUN collection to write the IntrinsicZernikes calibrations into. "
+            "e.g. LSSTCam/calib/DM-55048/intrinsicZernikes.v1.0/run.<date>"
```
erykoff, line 92: I don't know how the pipelines do this automatically, but really you want to have the code generate the timestamp.  But for this it would actually be fine I think.

python/lsst/ts/wep/bin/ingestIntrinsicZernikes.py
```diff
@@ -0,0 +1,190 @@
...
+        "--instrument",
+        type=str,
+        default="LSSTCam",
+        help="Instrument name. (default: LSSTCam)",
+        metavar="INST",
+    )
+    parser.add_argument(
+        "--input-collection",
+        type=str,
+        default="LSSTCam/aos/intrinsic",
```
jbkalmbach, line 65: Do we need to ask somebody to create this? We currently have `LSSTCam/runs/aos` but for outputs. Since this is an input I get why we would not want to put it there but we should consult with data curation folks.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -595,16 +591,33 @@ def run(
...
-            # then we only have one intrinsic table to use for both.
-            self.intrinsicMapExtra = self._createIntrinsicMap(intrinsicTables[0])
-            self.intrinsicMapIntra = self._createIntrinsicMap(intrinsicTables[0])
+            # then we only have one intrinsic calibration to use for both.
+            self.intrinsicZernikesExtra = intrinsicZernikes[0]
+            self.intrinsicZernikesIntra = intrinsicZernikes[0]
         else:
-            self.intrinsicMapExtra = self._createIntrinsicMap(intrinsicTables[0])
-            self.intrinsicMapIntra = self._createIntrinsicMap(intrinsicTables[1])
+            self.intrinsicZernikesExtra = intrinsicZernikes[0]
```
jbkalmbach: Not sure if there's a way to handle this or not but it would be nice to have a way to check that the first element in the list is actually extra and the second intra. The `runQuantum` and `lookup` function are ordered so this shouldn't happen in a pipeline but we will need to be careful running interactively.

Maybe `intrinsicZernikesExtra` and `intrinsicZernikesIntra` should be the inputs to the run method instead of hiding this here? Could do it with just a change in the `runQuantum` I think.

python/lsst/ts/wep/bin/ingestIntrinsicZernikes.py
gmegh: I will just remove the default, I don't think we ever intend to put anything on LSSTCam/aos/intrinsics

python/lsst/ts/wep/task/calcZernikesTask.py
gmegh: I am not sure about this, because then we need a special case for latiss to input None, no? Are we sure we want to do this in this PR?

jbkalmbach: I don't think it's too much. You just change `if len(intrinsicZernikes) == 0:` to check if they are both empty right?

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -545,15 +543,44 @@ def runQuantum(
...
+            intrinsicZernikesExtra = intrinsicZernikes[0]
+            intrinsicZernikesIntra = intrinsicZernikes[0]
+        else:
+            detectors = [ref.dataId["detector"] for ref in inputRefs.intrinsicZernikes]
+            if detectors[0] < detectors[1]:
+                intrinsicZernikes.reverse()
+            intrinsicZernikesIntra, intrinsicZernikesExtra = intrinsicZernikes
+
+        outputs = self.run(
+            **inputs,
```
jbkalmbach, line 575: Now I see why we popped it before. If you don't pop `intrinsicZernikes` then they will still be in inputs.

jbkalmbach (approved):
Looks good. Thanks!

PR #120, opened 2026-06-09
Added cut on recenter distance flags in donutStampSelectorTask.

jbkalmbach:


python/lsst/ts/wep/task/donutStampSelectorTask.py
```diff
@@ -298,9 +305,22 @@ def selectStamps(self, donutStamps: DonutStamps) -> pipeBase.Struct:
         elif self.config.selectWithMaxPowerGrad:
             self.log.warning("selectWithMaxPowerGrad==True but MAX_POWER_GRAD not in stamp metadata.")
 
+        # Select with recenter flags if requested
+        recenterFlagsSelect = np.ones(len(donutStamps), dtype="bool")
+        if self.config.selectWithRecenterFlags:
+            if "RECENTER_FLAGS" in list(donutStamps.metadata):
+                recenterFlags = np.asarray(donutStamps.metadata.getArray("RECENTER_FLAGS"))
```
elanaku: I think this line (or the line below) is causing an error and making the test fail. You may need an equivalent of the line above: fracBadPix[: len(fillVals)] = fillVals

elanaku (commented):
Looks good. It's just failing a test and I have some guess as to why...

python/lsst/ts/wep/task/donutStampSelectorTask.py
jbkalmbach: Thanks, made this take the same format as the ones above it and looks like it's passing that test on my machine.

elanaku (approved):
Looks good!

PR #125, opened 2026-06-17
Add triangle mode config

jbkalmbach:
Enable `ts_wep` to take advantage of danish v1.2 updates.

### DIFF
--- a/python/lsst/ts/wep/estimation/danish.py
+++ b/python/lsst/ts/wep/estimation/danish.py
@@ -84,6 +84,7 @@ class DanishAlgorithm(WfAlgorithm):
         bkgOrder: int = 0,
         doAoiThroughput: bool = False,
         systematicLossAlpha: float = 0.0,
+        triangleMode: bool = False,
     ) -> None:
         self.binning = binning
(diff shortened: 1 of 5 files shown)


python/lsst/ts/wep/estimation/danish.py
```diff
@@ -862,7 +882,11 @@ def _estimateZk(
         if self.doAoiThroughput:
             factory_kwargs["bandpass_filter"] = bandpass_filter
             factory_kwargs["airmass"] = airmass
-        factory = danish.DonutFactory(
+        if self.triangleMode:
+            factory_class = danish.DonutTriangleFactory
```
suberlak, line 906: Do you want to add testing whether `danish` has that method? That would prevent someone using an older danish from getting a value error...

suberlak (approved):
LGTM!

python/lsst/ts/wep/estimation/danish.py
jbkalmbach: Ok, added in some checks.

suberlak: Looks good!

suberlak (approved):

PR #126, opened 2026-06-19
Fixed issues with calibration intrinsics when running the unpaired da…

jbkalmbach:
…nish pipeline.

### DIFF
--- a/doc/news/RSO-660.bugfix.md
+++ b/doc/news/RSO-660.bugfix.md
@@ -1,0 +1,1 @@
+Fixed issues with calibration intrinsics when running the unpaired danish pipeline.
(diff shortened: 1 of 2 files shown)


suberlak (approved):
Looks good, does no test for calcZernikesUnpairedTask need updating ? I wonder why Merlin's CI caught it, and ts_wep unit tests didn't

jbkalmbach:
> Looks good, does no test for calcZernikesUnpairedTask need updating ? I wonder why Merlin's CI caught it, and ts_wep unit tests didn't

I'm looking now and I think it's because we never run the unpaired task as a pipeline in our tests so we didn't catch the `runQuantum` bug.

I'll push this and then open a ticket to make sure we have a way of testing that.

jbkalmbach:
https://rubinobs.atlassian.net/browse/RSO-661 is the related issue

PR #132, opened 2026-06-29
RSO-800: Added model images to always be included in zernikes table metadata f…

jbkalmbach:
…or danish estimates.


suberlak (approved):
Looks good, allows to always store model images to eg. pass on donut_viz in plkot_donut_fits. What's the time overhead for  always calling `model.model` ?

jbkalmbach:
Comparing timing before and after it seems very much the same so I think it's negligible.

PR #136, opened 2026-07-07
`Main`

jbkalmbach:


doc/version_history.rst
```diff
@@ -17,6 +17,16 @@ Version History
 
 .. towncrier release notes start
 
+17.7.0 (2026-07-07)
```
suberlak: Why is `v` missing from this line? (Others have `v17....` )

suberlak (approved):
Looks good, small comment (probably with towncrier run)

doc/version_history.rst
jbkalmbach: Fixed

PR #137, opened 2026-07-07
[RSO-810]: Add support for offsetting multiple optics in Batoid model

jbkalmbach:


python/lsst/ts/wep/instrument.py
```diff
@@ -675,22 +734,27 @@ def batoidOffsetOptic(self, value: str | None) -> None:
...
 
     @batoidOffsetValue.setter
-    def batoidOffsetValue(self, value: float | None) -> None:
-        """Set amount in meters the optic is offset in the batoid model.
+    def batoidOffsetValue(self, value: float | Sequence[float] | None) -> None:
+        """Set amount(s) in meters the optic(s) are offset in the batoid model.
 
-        This is the amount that batoidOffsetOptic is offset in the Batoid
+        This is the amount that each batoidOffsetOptic is offset in the Batoid
         model to calculate the equivalent detector offset for the model.
```
suberlak, line 811: So if you pass a list offsetting detector and camera,  what's the meaning of equivalent detector offset for detector offset provided ?

python/lsst/ts/wep/instrument.py
```diff
@@ -781,8 +849,7 @@ def _getIntrinsicZernikesCached(
         if defocalType is not None:
             defocalType = DefocalType(defocalType)
             defocalSign = +1 if defocalType == DefocalType.Extra else -1
```
suberlak, line 913: So does each element gain a different defocal type since you can move one in +ve  (eg detector) and another in -ve (camera) ?

python/lsst/ts/wep/instrument.py
```diff
@@ -631,42 +675,57 @@ def batoidModelName(self, value: str | None) -> None:
...
                 raise RuntimeError("There is no Batoid model set.")
-            elif not isinstance(value, str):
-                raise TypeError("batoidOffsetOptic must be a string or None.")
-            elif value not in self.getBatoidModel()._names:
-                raise ValueError(f"Optic {value} not found in the Batoid model.")
+            for optic in value:
+                if not isinstance(optic, str):
+                    raise TypeError("batoidOffsetOptic must be a string, a list of strings, or None.")
+                elif optic not in self.getBatoidModel()._names:
+                    raise ValueError(f"Optic {optic} not found in the Batoid model.")
```
suberlak, line 791: `batoidOffsetValue`'s setter explicitly clears caches (# Clear relevant caches), and `batoidModelName`'s setter clears _defocalOffsetBatoid. But the `batoidOffsetOptic` setter  doesn't show any cache clearing. Since `defocalOffset`, `_getIntrinsicZernikesCached`, and `_getIntrinsicZernikesTACached` all key off `self.batoidOffsetOptic`, changing which optic is offset after those caches are populated will silently return stale Zernikes/offsets computed for the old optic. Worth checking whether `clearCaches()` (or at least `_defocalOffsetBatoid = None + _getIntrinsicZernikesCached.cache_clear())` needs to be added here too

python/lsst/ts/wep/instrument.py
```diff
@@ -675,22 +734,27 @@ def batoidOffsetOptic(self, value: str | None) -> None:
...
     @batoidOffsetValue.setter
-    def batoidOffsetValue(self, value: float | None) -> None:
-        """Set amount in meters the optic is offset in the batoid model.
+    def batoidOffsetValue(self, value: float | Sequence[float] | None) -> None:
+        """Set amount(s) in meters the optic(s) are offset in the batoid model.
 
-        This is the amount that batoidOffsetOptic is offset in the Batoid
+        This is the amount that each batoidOffsetOptic is offset in the Batoid
         model to calculate the equivalent detector offset for the model.
         Note depending on the model, the sign of this value might matter.
```
suberlak, line 812: It would help to add concrete example, eg   "positive batoidOffsetValue with batoidOffsetOptic='LSSTCamera' corresponds to X mm of intra/extra motion, equivalent to Y mm at the detector." Even a one-line cross-reference to the FamCam config might help us debugging that in the future

tests/test_instrument.py
```diff
@@ -98,12 +98,49 @@ def testBadBatoidOffsetOptic(self) -> None:
...
+            Instrument(batoidOffsetOptic=["Detector", "fake"])
+        with self.assertRaises(TypeError):
+            Instrument(batoidOffsetOptic=["Detector", 1])
 
     def testBadBatoidOffsetValue(self) -> None:
         with self.assertRaises(RuntimeError):
             inst = Instrument()
             inst.batoidModelName = None
             inst.batoidOffsetValue = 1
+        # Mismatched lengths between optics and values are rejected
```
suberlak, line 112: ok, can you also validate what happens if someone sets `batoidOffsetOptic` and `batoidOffsetValue `separately post-construction in the wrong order (e.g. set a 2-element batoidOffsetOptic on an instrument that already has a 1-element batoidOffsetValue — does that raise error immediately?)

python/lsst/ts/wep/instrument.py
jbkalmbach: It refers to the `defocalOffset` above.

python/lsst/ts/wep/instrument.py
```diff
@@ -95,16 +95,19 @@ class Instrument:
...
+        accepted, or a list of strings to offset multiple optics
+        simultaneously (e.g. to model both a camera and detector piston for
+        Full Array Mode). (the default is None)
+    batoidOffsetValue : float, list of float, or None, optional
+        The value(s) in meters to offset the optic(s) in the Batoid model to
         calculate the equivalent detector offset for the model. The
-        detector offset is then used for everything else. Note that
-        depending on the model, the sign of this value might matter.
-        (the default is None)
+        detector offset is then used for everything else. If a list is
```
jmeyers314: do happen to know what "everything else" refers to here?

python/lsst/ts/wep/instrument.py
```diff
@@ -202,6 +205,18 @@ def checkConfig(self) -> None:
...
+        values = self.batoidOffsetValue
+        if values is not None and optics is not None:
+            nOptics = len(optics)
+            nValues = len(values)
+            if nOptics != nValues:
+                raise ValueError(
+                    "batoidOffsetOptic and batoidOffsetValue must have the same "
+                    f"length, but got {nOptics} optic(s) and {nValues} value(s)."
+                )
+
```
jmeyers314: Do we raise an error elsewhere if one, but not both of `values`, `optics` is None?

python/lsst/ts/wep/instrument.py
```diff
@@ -437,6 +449,38 @@ def offsetToZ4Defocus(self, offset: float) -> float:
...
         )[4]
 
+    def _applyBatoidOffsets(self, batoidModel: batoid.Optic, sign: int) -> batoid.Optic:
+        """Return a Batoid model with all configured optic offsets applied.
+
+        Each optic in ``batoidOffsetOptic`` is shifted along the optical axis
+        by ``sign * batoidOffsetValue[i]``. If ``batoidOffsetValue`` is not set
+        (i.e. the instrument specifies ``defocalOffset`` directly), each optic
+        is shifted by ``sign * defocalOffset`` instead, preserving the
+        single-offset behavior.
```
jmeyers314: Ah, I guess this is why `batoidOffsetValue` is not required even if `batoidOffsetOptic` is non-None?

python/lsst/ts/wep/instrument.py
```diff
@@ -437,6 +449,38 @@ def offsetToZ4Defocus(self, offset: float) -> float:
...
+        batoid.Optic
+            The Batoid model with all offsets applied.
+        """
+        optics = self.batoidOffsetOptic
+        if optics is None:
+            return batoidModel
+        values = self.batoidOffsetValue
+        if values is None:
+            # Fall back to shifting each optic by the scalar defocalOffset
+            values = [self.defocalOffset] * len(optics)
```
jmeyers314: Imagine len(batoidOffsetOptic) > 1 but batoidOffsetValue is None.  I think it this case it'd actually be less surprising if we raised an error rather than shifting multiple optics all the same amount.  Even for 3mm donuts, it's a choice for the FAM camera shift to equal the detector shift, not something fundamental.

jmeyers314 (commented):
Looks good!  A few small comments.

python/lsst/ts/wep/instrument.py
jbkalmbach: Yeah, I agree it's not asking much to specify each offset even if they are the same. Plus, it makes it easier to understand the shifts when just looking at the configuration yaml. I will remove this default "None" behavior.

python/lsst/ts/wep/instrument.py
jbkalmbach: Yeah, as mentioned below I will remove this option and make it required for both `batoidOffsetOptic` and `batoidOffsetOptic` to match in length.

python/lsst/ts/wep/instrument.py
jbkalmbach: Will modify this behavior so they need to be equal and raise the error here if one is `None` and the other is not.

python/lsst/ts/wep/instrument.py
jbkalmbach: I think this originally meant the equivalent detector offset was calculated from this value and that value was then applied to whatever optic was indicated by `batoidOffsetOptic`. This should be changed now.

python/lsst/ts/wep/instrument.py
jbkalmbach: Example added.

tests/test_instrument.py
jbkalmbach: I think this is all tested now with the new overhaul

python/lsst/ts/wep/instrument.py
jbkalmbach: They will all be relative to the defocal type specified (+ for detector and - for camera will move them in opposite directions)

python/lsst/ts/wep/instrument.py
```diff
@@ -353,34 +391,48 @@ def focalRatio(self) -> float:
...
+           equivalent Detector offset (cached after the first computation).
+
+        These two sources are mutually exclusive (enforced by checkConfig). A
+        ValueError is raised if neither is available.
+        """
         if self._defocalOffset is not None:
             return self._defocalOffset
         elif self._defocalOffsetBatoid is not None:
             return self._defocalOffsetBatoid
         elif self.batoidModelName is not None and self._batoidOffsetValue is not None:
```
suberlak: It looks like `defocalOffset` branch guards only on `_batoidOffsetValue is not None`, but
  `_applyBatoidOffsets` (line 538) falls back to `self.defocalOffset` when `_batoidOffsetOptic is None` — so if value is set while optic is None,
  reading `defocalOffset` recurses infinitely (RecursionError). Reproduced by: `inst.batoidOffsetOptic = None` on an LsstCam instrument, then
  reading `inst.defocalOffset`. Fix: also require `self._batoidOffsetOptic is not None` in this branch so it falls through to the clear ValueError
  instead of recursing. (Reverse case — optic set, value None — already raises ValueError, so the two guards are asymmetric.)

suberlak: Such as 
``` -        ValueError is raised if neither is available.
+        ValueError is raised if neither is available. Deriving from the Batoid
+        model requires both batoidOffsetOptic and batoidOffsetValue; if only
+        one of them is set (possible when they are cleared one at a time after
+        construction), a ValueError is raised rather than deriving a
+        half-specified offset.
         """
         if self._defocalOffset is not None:
             return self._defocalOffset
         elif self._defocalOffsetBatoid is not None:
             return self._defocalOffsetBatoid
-        elif self.batoidModelName is not None and self._batoidOffsetValue is not None:
+        elif (
+            self.batoidModelName is not None
+            and self._batoidOffsetOptic is not None
+            and self._batoidOffsetValue is not None
+        ):
```

suberlak (approved):
Looks good, one small comment that may prevent another failure mode

jmeyers314 (approved):
LGTM!

PR #138, opened 2026-07-15
DM-55506: Fix calcZernikesNeuralTask empty() OOD column crash

PetchMa:
- After RSO-663, `initZkTable()` always includes an average row, so `emptyZkTable["ood_score"] = []` raised `ValueError: Inconsistent data column lengths` whenever TARTS detected donuts but produced no usable Zernikes (including one-sided focus / empty-table cases).
- `empty()` now writes a length-1 NaN `ood_score` and marks the average row unused, matching Danish empty-table behavior.
- Also guard optional `None` model paths against `expandvars`, and update neural empty-path unit tests so they run against the current table convention.

Verified with unit tests and an offline on-sky pipetask replay of exposure `2026071300099`


jbkalmbach (approved):
Looks good. Thanks!

PR #139, opened 2026-07-17
Tickets/rso 826

jbkalmbach:


python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -670,6 +670,9 @@ def run(
         # Save the outputs in the table
         zkTable = self.createZkTable(zkCoeffRaw)
         zkTable.meta["estimatorInfo"] = zkCoeffRaw.wfEstInfo
+        zkTable.meta["estimatorInfo"]["algo"] = self.estimateZernikes.wfAlgoName.value
+        if "binning" in self.estimateZernikes.config.keys():
```
suberlak: Every other entry in `estimatorInfo (fwhm, caustic, fit_success, …)` is a per-donut list,
  but `algo` and `binning` are scalars. This is exactly what forced the special-casing on the donut_viz side (aggregate_visit.py setdefault, and the isinstance(value, 
  (list,tuple,ndarray)) filter in plot_aos_task.py). Any other user of `estimatorInfo` that assumes list-valued entries — e.g. iterating and indexing per donut, or
  np.array(value)[selected_rows] without the type guard — will now break or silently misbehave. I can't recall it there are any other users of this metadata,  but either consider switching that to lists as well (or len corresponding to number of donuts, which would make it possible not to have all the type-guarding), or a one-line comment here noting that
  these two keys are **intentionally** scalar.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -670,6 +670,9 @@ def run(
         # Save the outputs in the table
         zkTable = self.createZkTable(zkCoeffRaw)
         zkTable.meta["estimatorInfo"] = zkCoeffRaw.wfEstInfo
+        zkTable.meta["estimatorInfo"]["algo"] = self.estimateZernikes.wfAlgoName.value
```
suberlak: Line 672 binds the dict by reference, so lines 673-675 also add algo/binning back onto `zkCoeffRaw.wfEstInfo`. Harmless if zkCoeffRaw is discarded after this, but if it's
  ever reused or inspected downstream it now carries injected keys. A `dict(zkCoeffRaw.wfEstInfo`) copy on line 672 would make the ownership explicit.  But I see that we only use that as `outputZernikesRaw` , so that's probably okay.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -670,6 +670,9 @@ def run(
...
+        if "binning" in self.estimateZernikes.config.keys():
+            zkTable.meta["estimatorInfo"]["binning"] = self.estimateZernikes.config.binning
```
suberlak: Worth mentioning that we're storing here the configured binning, no necessarily what ran. `DanishAlgorithm` has a binning setter in l.144 in danish.py, so it would be basically a case if something changed how binning was set as opposed to just reading it from the config.

suberlak (approved):
Looks good, small comment-adding comments (mostly)

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: Moved to locate this in the estimation algorithms so every source gets a value now.

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: Same comment as above

PR #142, opened 2026-07-29
RSO-856: Load maskParams directly from danish.

jbkalmbach:


python/lsst/ts/wep/instrument.py
```diff
@@ -203,6 +216,7 @@ def copy(self) -> Instrument:
             batoidOffsetOptic=self.batoidOffsetOptic,
             batoidOffsetValue=self.batoidOffsetValue,
             maskParams=self.maskParams,
```
suberlak: copy() silently converts a file-backed mask into a frozen snapshot;  `maskParams=self.maskParams` loads from `maskParamsFile`. The copy receives that dict as an explicit `_maskParams`, which then takes precedence over maskParamsFile — so the copy is no longer file-backed:
 `orig  _maskParams is None : True`
 `copy  _maskParams is None : False`   

This means that any later `maskParamsFile` change on the copy is silently ignored, and `_logMaskVersions` misreports it as `"instrument-provided maskParams (overrides any danish file)".` (eg. in `tie.py:717` when binning > 1). It doesn't show up now  only because  every policy file has `maskParamsFile` commented out - i.e. it breaks exactly when this is uncommented  Perhaps it's better to pass  the raw stored value instead:  maskParams=getattr(self, "_maskParams", None),

python/lsst/ts/wep/instrument.py
```diff
@@ -121,7 +121,13 @@ class Instrument:
             to determine the center of the circle
             - radius: list of polynomial coeffs (in meters) for np.polyval()
             to determine the radius of the circle
+        If this is not None, it overrides any values set by maskParamsFile.
         None defaults to an empty dictionary.
```
suberlak: `None` now falls through to `maskParamsFil`e, or to the default pupil mask. The setter docstring was updated for this; also need to update this one

python/lsst/ts/wep/task/estimateZernikesBase.py
```diff
@@ -328,6 +330,71 @@ def estimateFromIndivStamps(
...
+            self.metadata["maskModel"] = maskModel
+        elif maskParamsFile is None:
+            self.log.info("Mask model: instrument-provided maskParams (no danish file)")
+            self.metadata["maskModel"] = maskModel
+        else:
+            try:
+                import danish
+
+                danishVersion = getattr(danish, "__version__", "unknown")
+                resolvedFile = os.path.basename(
```
suberlak: The provenance mechanism doesn't actually capture a version, but filename - te docstring says the resolved filename carries the version `"via the RubinObsc.yaml symlink"`, but it isn't a symlink in the installed danish so `os.path.realpath`  returns   basename: `RubinObsc.yaml`     (not `RubinObsc_v3.14_r_...`). So ` metadata["maskModel"]` is   `danish:RubinObsc.yaml` with no version in it. Perhaps recording `danish.__version__` in  the metadata too (like in line 367 for the log message) would help more in recording the danish version ?

suberlak (approved):
Looks good, just few small comments

python/lsst/ts/wep/task/estimateZernikesBase.py
jbkalmbach: Good idea. Did the same for batoid just for completeness.

